PR #35, opened 2026-02-25
feat: add kueue labels lint checks for all workload types

lburgazzoli:
Extend the kueue queue-name label validation to InferenceService,
LLMInferenceService, RayCluster, RayJob, and PyTorchJob workloads.
Checks verify that workloads with the kueue.x-k8s.io/queue-name label
are in kueue-enabled namespaces, and are only active when both the
owning component and Kueue are enabled on the DSC.

Also introduces shared validation logic in workloads/kueue/support.go,
AddCondition/SetConditions helpers on DiagnosticResult, and
GetLabel/HasLabel utilities in pkg/util/kube.

image

### DIFF
--- a/pkg/lint/checks/workloads/kueue/support.go
+++ b/pkg/lint/checks/workloads/kueue/support.go
@@ -1,0 +1,248 @@
+package kueue
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/opendatahub-io/odh-cli/pkg/constants"
+	"github.com/opendatahub-io/odh-cli/pkg/lint/check"
+	"github.com/opendatahub-io/odh-cli/pkg/lint/check/result"
+	"github.com/opendatahub-io/odh-cli/pkg/lint/check/validate"
+	"github.com/opendatahub-io/odh-cli/pkg/resources"
+	"github.com/opendatahub-io/odh-cli/pkg/util/client"
+	"github.com/opendatahub-io/odh-cli/pkg/util/components"
+	"github.com/opendatahub-io/odh-cli/pkg/util/kube"
+)
+
+// Messages for the "labeled in non-kueue namespace" condition.
+const (
+	MsgNoWorkloads        = "No %s instances found"
+	MsgNoLabeledWorkloads = "No %s(s) found with the kueue.x-k8s.io/queue-name label"
+	MsgAllValid           = "All %d %s(s) with the kueue.x-k8s.io/queue-name label are in kueue-enabled namespaces"
+	MsgNsNotKueueEnabled  = "Found %d %s(s) with the kueue.x-k8s.io/queue-name label in namespaces not enabled for kueue"
+)
+
+// Messages for the "missing label in kueue namespace" condition.
+const (
+	MsgNoWorkloadsInKueueNs  = "No %s(s) found in kueue-enabled namespaces"
+	MsgAllInKueueNsLabeled   = "All %d %s(s) in kueue-enabled namespaces have the kueue.x-k8s.io/queue-name label"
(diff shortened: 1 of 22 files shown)


pkg/lint/checks/workloads/kueue/support.go
```diff
@@ -0,0 +1,177 @@
...
+		// Find workloads that have the kueue queue label.
+		var labeled []types.NamespacedName
+
+		for _, item := range req.Items {
+			if kube.GetLabel(item, constants.LabelKueueQueueName) != "" {
+				labeled = append(labeled, types.NamespacedName{
+					Namespace: item.GetNamespace(),
+					Name:      item.GetName(),
+				})
+			}
```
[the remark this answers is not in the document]
lburgazzoli: fixed

pkg/lint/checks/workloads/kueue/support.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		allKueueNs, err := kueueEnabledNamespaces(ctx, req.Client)
+		if err != nil {
+			return err
+		}
+
+		kueueNamespaces := allKueueNs.Intersection(labeledNamespaces)
+
+		// Labeled workloads whose namespace is NOT kueue-enabled are impacted.
```
MarianMacik: Isn't the opposite what we want to check?

From the [fixing PR](https://github.com/opendatahub-io/opendatahub-operator/pull/3185/changes#diff-1c2bd1d24587a4323e59931edaa538b2d6538fbd63dad45ea40ef8d1d9bbd939R432):
```
// RHOAIENG-50667: Only when the Notebook's namespace is Kueue-labeled does the webhook require
// the queue-name label on the Notebook. Skip HWP migration only when namespace is Kueue-managed
// and the Notebook lacks the label; otherwise we would trigger webhook rejection and CrashLoopBackOff.
```

lburgazzoli: This is extending the checks to report additional possible misconfigurations:

- Workload with queue-name + namespace with kueue label -> GOOD
- Workload without queue-name + namespace with kueue label -> BAD
- Workload with queue-name + namespace without kueue label -> BAD
- Workload without queue-name + namespace without kueue label -> GOOD

MarianMacik: Before it was iterating over namespaces which have labelled workloads and checking if they are found in intersected namespaces, adding if they were not, so what was left were namespaces which don't have labels but workloads have. So just one use case was captured.

lburgazzoli: Yeah, I did try to capture all the possible cases, given some are probably not actually really possible unless you are artificially creating such scenario

MarianMacik: I was trying to catch the root cause which happened in that blocker (workloads without label in labeled namespace), so it was possible. Looks good now.

andresllh:
tested isvcs and the tool was able to recognize a misconfigured isvc
image

andresllh (approved):
works for ISVCs

andyatmiami (approved):
/lgtm

changes in line (and in a lot of cases **better**) with what I originally delivered with notebooks.

confirmed reporting on notebooks still functionally correct:

Image

---

❓ While not needing to _hold up this PR_ - I would question personally whether **this new detection case** is really a `critical` (blocking) impact:
- `Found 1 Notebook(s) with the kueue.x-k8s.io/queue-name label in namespaces not enabled for kueue`

It is **certainly** a _misconfiguration_ worth flagging - but imho a _lesser severity_ as it doesn't impede upgrade scenario from progressing.. i.e. this situation is ignored in 2.25 and would continue to be ignored during upgrade to 3.3...  So its clearly **not good** - but I could see it being moreso a `warning`..

BUUUUT as I say that.. I realize for other components like `Ray` - this **IS** a more impactful check.. as it used to work **without** the `namespace` label but now the namespace is required.

So its probably fine to just leave this as `critical` and not overanalyze (like I am doing right now).  The alternative of making the severity of this check _component-specific_ seems like overkill...

lburgazzoli:
@laurafitzgerald can you test it vs ray ?

valdar (approved):

PR #55, opened 2026-03-24
fix: reduce false positives for ServiceMesh capability conditions in DSCI check

lburgazzoli:
Split the single isRemovedCapabilityServiceMesh helper into two separate
functions with different exemption logic per condition type:

- CapabilityServiceMesh: False with reason Removed or Unmanaged is treated
  as a healthy intentional state (not flagged).
- CapabilityServiceMeshAuthorization: only False with reason NotReady is a
  genuine problem; all other False reasons (Removed, Unmanaged,
  MissingOperator) are non-blocking and no longer flagged.

Made-with: Cursor

## Testing

## Review checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined readiness detection: Service Mesh False with reasons Removed or Unmanaged is treated non-blocking; Service Mesh Authorization False is considered blocking only when reason is NotReady, yielding clearer readiness/impact results.
* **Tests**
  * Added parameterized readiness tests covering capability types and reason combinations, validating readiness condition, message prefixes, and impact classification.

### DIFF
--- a/pkg/lint/checks/platform/dscinitialization/dscinitialization_readiness_test.go
+++ b/pkg/lint/checks/platform/dscinitialization/dscinitialization_readiness_test.go
@@ -238,44 +238,139 @@ func TestDSCInitializationReadinessCheck_NotReadyWithUnhappyConditions(t *testin
 	g.Expect(result.Status.Conditions[0].Impact).To(Equal(resultpkg.ImpactBlocking))
 }
 
-func TestDSCInitializationReadinessCheck_ReadyWithRemovedCapabilities(t *testing.T) {
-	g := NewWithT(t)
-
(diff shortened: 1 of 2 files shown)


pkg/lint/checks/platform/dscinitialization/dscinitialization_readiness_test.go
MarianMacik: This test can be removed now so it is not duplicated.

MarianMacik (approved):

pkg/lint/checks/platform/dscinitialization/dscinitialization_readiness_test.go
lburgazzoli: done

PR #58, opened 2026-03-30
feat: add Service Mesh v3 dependency check for 2.x to 3.x upgrades

csams:
## Description

Validates Red Hat Service Mesh v3 operator required by OpenShift Ingress in disconnected environments is available in the redhat-operators catalog source when upgrading from RHOAI 2.x to 3.x. Includes guidance for disconnected installs to mirror appropriate images.

## Testing

- [X] Unit tests pass (`make test`)
- [X] Linter passes (`make lint`)
- [X] New functionality includes tests

## Review checklist

## Summary by CodeRabbit

* **New Features**
  * Added a dependency lint check that validates Service Mesh 2.x→3.x upgrade readiness and provides actionable remediation.

* **Tests**
  * Added comprehensive tests for version/channel detection, permission failures, missing metadata, and diagnostic messaging.

* **Documentation**
  * Updated lint docs to reflect the added dependency registration.

* **Chores**
  * Added support for PackageManifest resources used by the new check.

### DIFF
--- a/pkg/lint/checks/dependencies/servicemesh/servicemesh.go
+++ b/pkg/lint/checks/dependencies/servicemesh/servicemesh.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -58,6 +59,7 @@ func (c *Check) Validate(ctx context.Context, target check.Target) (*result.Diag
 			metav1.ConditionFalse,
 			check.WithReason(check.ReasonResourceNotFound),
 			check.WithMessage("ingress-operator deployment not found in openshift-ingress-operator namespace"),
+			check.WithRemediation("Check that the openshift-ingress-operator namespace and the ingress-operator deployment exist in the cluster."),
 			check.WithImpact(result.ImpactBlocking),
 		))
 
@@ -70,6 +72,7 @@ func (c *Check) Validate(ctx context.Context, target check.Target) (*result.Diag
 			metav1.ConditionFalse,
 			check.WithReason(check.ReasonInsufficientData),
 			check.WithMessage("Unable to read ingress-operator deployment (insufficient permissions)"),
+			check.WithRemediation("Grant read access to deployments in the openshift-ingress-operator namespace."),
 			check.WithImpact(result.ImpactBlocking),
 		))
 
@@ -87,6 +90,7 @@ func (c *Check) Validate(ctx context.Context, target check.Target) (*result.Diag
 			metav1.ConditionFalse,
 			check.WithReason(check.ReasonDependencyUnavailable),
 			check.WithMessage("GATEWAY_API_OPERATOR_VERSION env var not found on ingress-operator deployment"),
+			check.WithRemediation("Verify the ingress-operator deployment in the openshift-ingress-operator namespace has the GATEWAY_API_OPERATOR_VERSION environment variable."),
 			check.WithImpact(result.ImpactBlocking),
 		))
 
@@ -99,16 +103,21 @@ func (c *Check) Validate(ctx context.Context, target check.Target) (*result.Diag
 			metav1.ConditionFalse,
 			check.WithReason(check.ReasonDependencyUnavailable),
 			check.WithMessage("GATEWAY_API_OPERATOR_VERSION env var is empty on ingress-operator deployment"),
+			check.WithRemediation("Verify the ingress-operator deployment in the openshift-ingress-operator namespace has a non-empty GATEWAY_API_OPERATOR_VERSION environment variable."),
 			check.WithImpact(result.ImpactBlocking),
 		))
 
 		return dr, nil
 	}
 
-	// Step 3: Build the expected CSV name.
-	requiredCSV := "servicemeshoperator3.v" + requiredVersion
+	// Step 3: The env var contains the full CSV name (e.g. "servicemeshoperator3.v3.1.0").
+	requiredCSV := requiredVersion
+	displayVersion := strings.TrimPrefix(requiredCSV, "servicemeshoperator3.v")
 
 	// Step 4: Find the servicemeshoperator3 PackageManifest from redhat-operators in openshift-marketplace.
+	// Multiple catalog sources can produce PackageManifests with the same name, so we list all
+	// PackageManifests and filter by .status.catalogSource. A direct get-by-name would return a
+	// non-deterministic result when multiple catalog sources provide the same package.
 	manifests, err := client.List[*unstructured.Unstructured](ctx, target.Client, resources.PackageManifest,
 		func(pm *unstructured.Unstructured) (bool, error) {
 			if pm.GetName() != "servicemeshoperator3" || pm.GetNamespace() != "openshift-marketplace" {
@@ -132,6 +141,7 @@ func (c *Check) Validate(ctx context.Context, target check.Target) (*result.Diag
 			metav1.ConditionFalse,
 			check.WithReason(check.ReasonResourceNotFound),
 			check.WithMessage("servicemeshoperator3 PackageManifest from redhat-operators not found in openshift-marketplace"),
+			check.WithRemediation("Mirror servicemeshoperator3 into the redhat-operators catalog source in the openshift-marketplace namespace."),
 			check.WithImpact(result.ImpactBlocking),
 		))
 
@@ -141,7 +151,7 @@ func (c *Check) Validate(ctx context.Context, target check.Target) (*result.Diag
 	pm := manifests[0]
 
 	// Step 5: Extract available CSVs from all channels.
-	availableCSVs, err := jq.Query[[]string](pm, "[.status.channels[]?.currentCSV]")
+	availableCSVs, err := jq.Query[[]string](pm, "[.status.channels[]?.entries[]?.name]")
 	if err != nil {
 		return nil, fmt.Errorf("querying available CSVs: %w", err)
 	}
@@ -152,14 +162,14 @@ func (c *Check) Validate(ctx context.Context, target check.Target) (*result.Diag
 			check.ConditionTypeAvailable,
 			metav1.ConditionTrue,
 			check.WithReason(check.ReasonResourceFound),
-			check.WithMessage("%s version %s is available in the cluster catalog", displayName, requiredVersion),
+			check.WithMessage("%s (%s) is available in the 'redhat-operators' cluster catalog", displayName, requiredCSV),
 		))
 	} else {
 		dr.SetCondition(check.NewCondition(
 			check.ConditionTypeAvailable,
 			metav1.ConditionFalse,
 			check.WithReason(check.ReasonDependencyUnavailable),
-			check.WithMessage("%s version %s is not available in the cluster catalog", displayName, requiredVersion),
+			check.WithMessage("%s version %s is not available in the cluster catalog", displayName, displayVersion),
 			check.WithRemediation(fmt.Sprintf("Mirror %s into your environment. It must be available in the redhat-operators catalog source in the openshift-marketplace namespace. See the pre-requisite instructions in the RHOAI 2.x to 3.x upgrade guide.", requiredCSV)),
 			check.WithImpact(result.ImpactBlocking),
 		))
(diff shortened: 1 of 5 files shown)


pkg/lint/checks/dependencies/servicemesh/servicemesh.go
andyatmiami, line 1: For the following `check.NewCondition` returns that are `ConditionFalse` (i.e. failures) - I think we want to use a combination of `WithMessage` and `WithRemediation` - whereas in the codebase today only `WithMessage` is used:
- https://github.com/opendatahub-io/odh-cli/pull/58/changes#diff-9bd87bc2a154359b6f6e0734dfac48df6869a0b47232f3815b5ff84225c23fb5R58
- https://github.com/opendatahub-io/odh-cli/pull/58/changes#diff-9bd87bc2a154359b6f6e0734dfac48df6869a0b47232f3815b5ff84225c23fb5R70
- https://github.com/opendatahub-io/odh-cli/pull/58/changes#diff-9bd87bc2a154359b6f6e0734dfac48df6869a0b47232f3815b5ff84225c23fb5R81

**Message** describes what the problem is — it's the diagnostic finding. It answers "what did the check observe?"

**Remediation** describes how to fix it — it's the actionable guidance. It answers "what should the user do?"

This matters because they are separate fields in the structured output. The Condition struct (pkg/lint/check/result/result.go:60-70) serializes them independently:

```
  {
    "message": "Red Hat Service Mesh v3 is not installed",
    "impact": "blocking",
    "remediation": "Install Red Hat Service Mesh v3 operator (>= 3.1.0) from OperatorHub..."
  }
```

  When remediation is embedded in the message, consumers of the JSON/YAML output (automation, dashboards, support tooling) can't programmatically distinguish the diagnosis from the fix guidance — they're mushed into a single string. The remediation field will be empty/missing in the structured output, which breaks the contract other checks maintain.

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/opendatahub-io/odh-cli/pkg/lint/check/result"
+	"github.com/opendatahub-io/odh-cli/pkg/lint/check/validate"
+	"github.com/opendatahub-io/odh-cli/pkg/util/version"
+)
+
+const kind = "servicemesh-v3"
+
+const displayName = "Red Hat Service Mesh v3"
+
+const disconnectedNote = "If performing a disconnected install, ensure the appropriate images are mirrored into your environment."
```
andyatmiami: We had also discussed potentially passing a `--disconnected` flag here for more targeted behavior... but I think just baking this into the message on any/all cluster `lint` check is probably fine for the following reasons:

- No great/simple/deterministic way to know a cluster is operated in `disconnected` mode
- as this is the **ONLY** "disconnected-specific" behavior - if a client _forgot_ to pass this flag... then they miss this important note
- no great/simple/deterministic way to know if the image was actually mirrored appropriately - so "vague-but-accurate" messaging is the only thing we can do

No request for change here - just surfacing my thoughts on why this existing code is (imho) as suitable an implementation as any other.

pkg/lint/command.go
```diff
@@ -104,9 +105,10 @@ func NewCommand(
 	registry.MustRegister(modelmesh.NewRemovalCheck())
 	registry.MustRegister(trainingoperator.NewDeprecationCheck())
 
-	// Dependencies (2)
+	// Dependencies (3)
 	registry.MustRegister(certmanager.NewCheck())
 	registry.MustRegister(openshift.NewCheck())
+	registry.MustRegister(servicemesh.NewCheck())
```
andyatmiami, line 111: Registering this check here means we are now formally elevating the OSSM v3 to an explicit pre-req on that upgrade process - whereas before we had been implicitly relying on the RHOAI operator upgrade to "pull this in".

I am not claiming this needs to be changed - but simply want to explicitly callout the _potential ramifications_ of this decision in case docs/etc need to be adjusted in light of this.

andyatmiami (changes requested):
Thanks for getting this raised @csams ...

My only real _request for change_ is around usage of `WithRemediation` - the other comments on this PR are simply moreso for historical documentation/awareness.

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
```diff
@@ -0,0 +1,169 @@
...
+			check.WithImpact(result.ImpactBlocking),
+		))
+
+		return dr, nil
+	}
+
+	pm := manifests[0]
+
+	// Step 5: Extract available CSVs from all channels.
+	availableCSVs, err := jq.Query[[]string](pm, "[.status.channels[]?.currentCSV]")
```
asmigala: I think this should rather be 
```
.status.channels[].entries[].version
```

asmigala (changes requested):

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
```diff
@@ -0,0 +1,169 @@
...
+			catalogSource, err := jq.Query[string](pm, ".status.catalogSource")
+			if err != nil {
+				return false, nil
+			}
+
+			return catalogSource == "redhat-operators", nil
+		})
+	if err != nil {
+		return nil, fmt.Errorf("listing PackageManifests: %w", err)
+	}
```
asmigala, line 154: With @MarianMacik we found an interesting thing about packagemanifests: When you have multiple catalogsources containing the same package, the API will return multiple packagemanifest resources with the same name.

My understanding is that this code should actually handle that - we go through all package manifests and filter those that are called `servicemeshoperator3` and have `spec.catalogSource` set to `redhat-operators`, so any other packagemanifests with the same name but non-matching catalog source will be ignored.

Just making sure I understand correctly. 

Also, could you add a comment why it's done this way, so that in case of any refactoring, the code is not switched to getting the packagemanifest by name  (which appears to return a random one).

asmigala (changes requested):

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
asmigala: Alternatively, we can list packagemanifests with the label `catalog: redhat-operators` and try to find `servicemeshoperator3` in that

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil, fmt.Errorf("querying available CSVs: %w", err)
+	}
+
+	// Step 6: Check if the required CSV is available.
+	if slices.Contains(availableCSVs, requiredCSV) {
+		dr.SetCondition(check.NewCondition(
+			check.ConditionTypeAvailable,
+			metav1.ConditionTrue,
+			check.WithReason(check.ReasonResourceFound),
+			check.WithMessage("%s version %s is available in the cluster catalog", displayName, requiredVersion),
```
MarianMacik:
```suggestion
			check.WithMessage("%s version %s is available in the 'redhat-operators' cluster catalog", displayName, requiredVersion),
```

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
MarianMacik: Correct, this is probably more robust (querying based on `.status.catalogSource`)  than using labels.

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
```diff
@@ -0,0 +1,169 @@
...
+			check.WithReason(check.ReasonDependencyUnavailable),
+			check.WithMessage("GATEWAY_API_OPERATOR_VERSION env var is empty on ingress-operator deployment"),
+			check.WithImpact(result.ImpactBlocking),
+		))
+
+		return dr, nil
+	}
+
+	// Step 3: Build the expected CSV name.
+	requiredCSV := "servicemeshoperator3.v" + requiredVersion
```
asmigala: The env var we extracted this from is already in the form `${package}.{version}`, so this won't work. Also with my other suggestion to use `.status.channels[].entries[].name` to get _all_ CSVs (which is necessary on connected clusters), we can use the `requiredVersion` directly

asmigala (changes requested):

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
MarianMacik: As I said [here](https://docs.google.com/document/d/1UCy1BNaXbZ0pKsZ_vHqcQAeIZi4Px2hyStSVBqeEgJM/edit?disco=AAAB2yVNnt4) it may happen that newer versions have been mirrored as well, so we should check all versions, not just currentCSV, which references the latest one for the channel.
Or as we discussed with @asmigala we can check full CSV names by doing
.status.channels[].entries[].name which will report all available CSVs. Then we can compare directly env var value with values from package manifest.

asmigala: Actually 
```
.status.channels[].entries[].name
```
will give the CSV, which will make things easier since the env var on the CIO deployment is also the full CSV

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
```diff
@@ -0,0 +1,169 @@
...
+type Check struct {
+	check.BaseCheck
+}
+
+func NewCheck() *Check {
+	return &Check{
+		BaseCheck: check.BaseCheck{
+			CheckGroup:       check.GroupDependency,
+			Kind:             kind,
+			Type:             check.CheckTypeInstalled,
```
asmigala, line 36: Nit: this is not really about installation, but I see we don't have a better check type currenlty

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
```diff
@@ -0,0 +1,169 @@
...
+			check.WithReason(check.ReasonResourceFound),
+			check.WithMessage("%s version %s is available in the cluster catalog", displayName, requiredVersion),
+		))
+	} else {
+		dr.SetCondition(check.NewCondition(
+			check.ConditionTypeAvailable,
+			metav1.ConditionFalse,
+			check.WithReason(check.ReasonDependencyUnavailable),
+			check.WithMessage("%s version %s is not available in the cluster catalog", displayName, requiredVersion),
+			check.WithRemediation(fmt.Sprintf("Mirror %s into your environment. It must be available in the redhat-operators catalog source in the openshift-marketplace namespace. See the pre-requisite instructions in the RHOAI 2.x to 3.x upgrade guide.", requiredCSV)),
```
apodhrad: > Make sure %s is in the redhat-operators catalog source

apodhrad: Note that we don't know what was mirrored and what was not as it lives in external service which is fully under customer's control.

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
```diff
@@ -0,0 +1,179 @@
...
+		return dr, nil
+	}
+
+	pm := manifests[0]
+
+	// Step 5: Extract available CSVs from all channels.
+	availableCSVs, err := jq.Query[[]string](pm, "[.status.channels[]?.entries[]?.name]")
+	if err != nil {
+		return nil, fmt.Errorf("querying available CSVs: %w", err)
+	}
```
MarianMacik: Is `?` needed if we are anyway checking for errors?

pkg/lint/checks/dependencies/servicemesh/servicemesh.go
```diff
@@ -0,0 +1,179 @@
...
+			check.WithRemediation("Grant read access to deployments in the openshift-ingress-operator namespace."),
+			check.WithImpact(result.ImpactBlocking),
+		))
+
+		return dr, nil
+	}
+
+	// Step 2: Extract the required version from the GATEWAY_API_OPERATOR_VERSION env var.
+	requiredVersion, err := jq.Query[string](deploy,
+		`[.spec.template.spec.containers[].env[]? | select(.name == "GATEWAY_API_OPERATOR_VERSION") | .value] | first`)
```
MarianMacik: As @asmigala correctly pointed out, to be safe, we should also check GATEWAY_API_OPERATOR_CHANNEL
and then we need to find `.status.channels[] | select (.name == "<channel>") | .entries[].name`

I.e. the whole command is `oc get packagemanifest -o json | jq '.items[] | select (.metadata.name == "servicemeshoperator3" and .status.catalogSource == "redhat-operators") | .status.channels[] | select (.name == "<channel>") | .entries[].name'` which is what basically this check would be doing.

MarianMacik (approved):

PR #62, opened 2026-04-07
chore: updates to target 3.3.2 release

andyatmiami:
## Description

for reasons not worth explicitly outlining in this PR description - the version of RHOAI we are targeting for migration is now 3.3.2 (when previously it was 3.3.1).

While CanApply() checks - and the majority of string references - are generally only gated on 3.3 - there were a handful of references in user-facing strings and unit tests that still explicitly listed 3.3.1.

For sake of ease and consistency - this PR does a wholesale replace of the "3.3.1" string with "3.3.2" - even in cases where it was "just" a unit test and didn't strictly speaking require changes.
- This was an old fashioned "global search and replace" - no AI used in construction of this change set :smile:

## Testing

## Review checklist

### DIFF
--- a/pkg/lint/checks/components/kueue/kueue.go
+++ b/pkg/lint/checks/components/kueue/kueue.go
@@ -21,14 +21,14 @@ const (
 	checkTypeManagementState = "management-state"
 
 	// Deferred: parameterize hardcoded version references using ComponentRequest.TargetVersion.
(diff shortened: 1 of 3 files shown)


andyatmiami:
## Verification

Changes restricted to string manipulations and unit test files.. but out of an **abundance of caution** - I did run this against the "big/bad BU cluster" to ensure no regressions (which unsurprisingly succeeded!)

image

csams (approved):

PR #68, opened 2026-04-27
feat: add install, and graph subcommands for deps command.

ShettyGaurav:
## Description

Adds `install` and `graph` subcommands to the `deps` command for managing ODH/RHOAI operator dependencies via OLM.

### New Commands

**`kubectl odh deps install [DEPENDENCY]`**
- Creates namespace, OperatorGroup, and OLM Subscription for missing dependencies
- Waits for CSV to reach Succeeded phase
- Supports `--dry-run` to preview resources without applying
- Supports `--include-optional` to install optional (enabled=auto) dependencies
- Supports `--timeout` for CSV wait duration (default 5m)
- Labels created resources with `app.kubernetes.io/managed-by: odh-cli`
- Warns if existing Subscription has different channel/source
- Fails fast on API errors during CSV wait (instead of timing out silently)

**`kubectl odh deps graph`**
- Displays dependency relationships visually
- Shows which components require each operator
- Shows transitive dependencies between operators

### Backward Compatibility

Running `kubectl odh deps` without a subcommand still executes the list behavior (shows dependency status).

### Files Changed

- `cmd/deps/deps.go` - Refactored to support subcommands
- `pkg/deps/manifest.go` - Extended with fields for install support (Config, TargetNamespaces, Source)
- `pkg/deps/install.go` - Install command implementation
- `pkg/deps/graph.go` - Graph command implementation
- `pkg/deps/*_test.go` - Unit tests for new commands

## Testing

## Review checklist

## Summary by CodeRabbit

* **New Features**
  * Added top-level dependency manager with new subcommands: install (with --dry-run, target selection, --timeout, --include-optional) and graph (with --refresh), plus updated help/examples.
  * Graph output: human-readable dependency relationships and required/depends-on lists.
  * Dry-run prints install YAML; live install reports per-dependency results and waits for readiness.

* **Bug Fixes**
  * Improved version detection by falling back to cluster-wide CSV lookup when needed.

* **Tests**
  * Added tests covering graph, install, matching logic, dry-run output, and install decisioning.

### DIFF
--- a/cmd/deps/deps.go
+++ b/cmd/deps/deps.go
@@ -6,41 +6,75 @@ import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/genericiooptions"
 
+	"github.com/opendatahub-io/odh-cli/pkg/cmd"
 	depspkg "github.com/opendatahub-io/odh-cli/pkg/deps"
 	clierrors "github.com/opendatahub-io/odh-cli/pkg/util/errors"
 )
 
 const (
 	cmdName  = "deps"
-	cmdShort = "Show operator dependencies for ODH/RHOAI"
+	cmdShort = "Manage operator dependencies for ODH/RHOAI"
 )
 
 const cmdLong = `
-Shows operator dependencies required by ODH/RHOAI components.
+Manage operator dependencies required by ODH/RHOAI components.
 
-Displays dependency status by querying OLM subscriptions on the cluster.
+Without a subcommand, displays dependency status by querying OLM subscriptions.
 Each dependency shows:
   - Installation status (installed, missing, optional)
   - Installed version (if available)
   - Namespace where the operator runs
   - Components that require this dependency
 
-Examples:
+Subcommands:
+  install     Install missing operator dependencies via OLM
+  graph       Show dependency relationships
+`
+
+const cmdExample = `
   # Show all dependencies
   kubectl odh deps
 
   # Show dependencies for a specific version
   kubectl odh deps --version 3.4.0
 
-  # Refresh manifest from odh-gitops (fetch latest)
-  kubectl odh deps --refresh
-
   # Output as JSON
   kubectl odh deps -o json
 
   # Dry run (show manifest data without cluster query)
   kubectl odh deps --dry-run
+
+  # Install all missing required dependencies
+  kubectl odh deps install
+
+  # Install a specific dependency
+  kubectl odh deps install cert-manager
+
+  # Show what would be installed without executing
+  kubectl odh deps install --dry-run
+
+  # Show dependency graph
+  kubectl odh deps graph
 `
+
+// runCommand executes the Complete/Validate/Run lifecycle with error handling.
+//
+//nolint:wrapcheck // HandleError returns an already-handled error
+func runCommand(cobraCmd *cobra.Command, c cmd.Command, outputFormat string) error {
+	if err := c.Complete(); err != nil {
+		return clierrors.HandleError(cobraCmd, err, outputFormat)
+	}
+
+	if err := c.Validate(); err != nil {
+		return clierrors.HandleError(cobraCmd, err, outputFormat)
+	}
+
+	if err := c.Run(cobraCmd.Context()); err != nil {
+		return clierrors.HandleError(cobraCmd, err, outputFormat)
+	}
+
+	return nil
+}
 
 // AddCommand adds the deps command to the root command.
 func AddCommand(root *cobra.Command, flags *genericclioptions.ConfigFlags) {
@@ -50,52 +84,86 @@ func AddCommand(root *cobra.Command, flags *genericclioptions.ConfigFlags) {
 		ErrOut: root.ErrOrStderr(),
 	}
 
-	command := depspkg.NewCommand(streams, flags)
+	// Default list command (backward compatibility)
+	listCommand := depspkg.NewCommand(streams, flags)
 
-	cmd := &cobra.Command{
+	depsCmd := &cobra.Command{
 		Use:           cmdName,
 		Short:         cmdShort,
 		Long:          cmdLong,
+		Example:       cmdExample,
 		SilenceUsage:  true,
 		SilenceErrors: true,
-		RunE: func(cmd *cobra.Command, _ []string) error {
-			outputFormat := command.Output
-
-			if err := command.Complete(); err != nil {
-				if clierrors.WriteStructuredError(cmd.ErrOrStderr(), err, outputFormat) {
-					return clierrors.NewAlreadyHandledError(err)
-				}
-
-				clierrors.WriteTextError(cmd.ErrOrStderr(), err)
-
-				return clierrors.NewAlreadyHandledError(err)
-			}
-
-			if err := command.Validate(); err != nil {
-				if clierrors.WriteStructuredError(cmd.ErrOrStderr(), err, outputFormat) {
-					return clierrors.NewAlreadyHandledError(err)
-				}
-
-				clierrors.WriteTextError(cmd.ErrOrStderr(), err)
-
-				return clierrors.NewAlreadyHandledError(err)
-			}
-
-			if err := command.Run(cmd.Context()); err != nil {
-				if clierrors.WriteStructuredError(cmd.ErrOrStderr(), err, outputFormat) {
-					return clierrors.NewAlreadyHandledError(err)
-				}
-
-				clierrors.WriteTextError(cmd.ErrOrStderr(), err)
-
-				return clierrors.NewAlreadyHandledError(err)
-			}
-
-			return nil
+		RunE: func(cobraCmd *cobra.Command, _ []string) error {
+			return runCommand(cobraCmd, listCommand, listCommand.Output)
 		},
 	}
 
-	command.AddFlags(cmd.Flags())
+	listCommand.AddFlags(depsCmd.Flags())
 
-	root.AddCommand(cmd)
+	// Add subcommands
+	addInstallCommand(depsCmd, flags, streams)
+	addGraphCommand(depsCmd, flags, streams)
+
+	root.AddCommand(depsCmd)
+}
+
+// addInstallCommand adds the install subcommand.
+// Note: This command is text-only (progress messages) and does not support --output flag.
+func addInstallCommand(parent *cobra.Command, flags *genericclioptions.ConfigFlags, streams genericiooptions.IOStreams) {
+	installCommand := depspkg.NewInstallCommand(streams, flags)
+
+	installCmd := &cobra.Command{
+		Use:   "install [DEPENDENCY]",
+		Short: "Install missing operator dependencies via OLM",
+		Long: `Install missing operator dependencies required by ODH/RHOAI.
+
+Creates namespace, OperatorGroup, and OLM Subscription for each missing dependency.
+Waits for the CSV (ClusterServiceVersion) to reach Succeeded phase.
+
+If a dependency name is provided, only that dependency is installed.
+Otherwise, all missing required dependencies are installed.
+
+Note: This command outputs progress text only; the parent's --output flag does not apply.`,
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		Args:          cobra.MaximumNArgs(1),
+		RunE: func(cobraCmd *cobra.Command, args []string) error {
+			installCommand.TargetDep = ""
+			if len(args) > 0 {
+				installCommand.TargetDep = args[0]
+			}
+
+			return runCommand(cobraCmd, installCommand, "")
+		},
+	}
+
+	installCommand.AddFlags(installCmd.Flags())
+	parent.AddCommand(installCmd)
+}
+
+// addGraphCommand adds the graph subcommand.
+// Note: This command is text-only (ASCII visualization) and does not support --output flag.
+func addGraphCommand(parent *cobra.Command, flags *genericclioptions.ConfigFlags, streams genericiooptions.IOStreams) {
+	graphCommand := depspkg.NewGraphCommand(streams, flags)
+
+	graphCmd := &cobra.Command{
+		Use:   "graph",
+		Short: "Show dependency relationships visually",
+		Long: `Display dependency relationships as a formatted list.
+
+Shows which components require which operators, and which operators
+depend on other operators (transitive dependencies).
+
+Note: This command outputs text visualization only; the parent's --output flag does not apply.`,
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		Args:          cobra.NoArgs,
+		RunE: func(cobraCmd *cobra.Command, _ []string) error {
+			return runCommand(cobraCmd, graphCommand, "")
+		},
+	}
+
+	graphCommand.AddFlags(graphCmd.Flags())
+	parent.AddCommand(graphCmd)
 }
--- a/pkg/deps/install.go
+++ b/pkg/deps/install.go
@@ -1,0 +1,1032 @@
+package deps
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+	"time"
+
+	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/spf13/pflag"
+	"golang.org/x/sync/errgroup"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+
+	"github.com/opendatahub-io/odh-cli/pkg/cmd"
+	"github.com/opendatahub-io/odh-cli/pkg/util/client"
+	utilserrors "github.com/opendatahub-io/odh-cli/pkg/util/errors"
+	"github.com/opendatahub-io/odh-cli/pkg/util/iostreams"
+)
+
+// Verify InstallCommand implements cmd.Command interface at compile time.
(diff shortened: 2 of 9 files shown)


ShettyGaurav:
@coderabbitai review

pkg/deps/install.go
```diff
@@ -0,0 +1,774 @@
...
+
+	if err := c.createSubscription(ctx, dep); err != nil {
+		result.Status = InstallStatusFailed
+		result.Error = err
+		c.printFailure(w, dep.DisplayName, err)
+
+		return result
+	}
+
+	_, _ = fmt.Fprintln(w, msgCreatingSubscription)
```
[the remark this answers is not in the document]
den-rgb: +1

pkg/deps/install.go
```diff
@@ -0,0 +1,774 @@
...
+	source := dep.Source
+	if source == "" {
+		source = defaultCatalogSource
+	}
+
+	sub := &operatorsv1alpha1.Subscription{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "operators.coreos.com/v1alpha1",
+			Kind:       "Subscription",
+		},
```
den-rgb, line 705: add Labels: map[string]string{labelManagedBy: labelManagedByValue} to the Subscription ObjectMeta. to match Namespace and OperatorGroup

pkg/deps/install.go
```diff
@@ -0,0 +1,774 @@
...
+				return "", result.err
+			}
+
+			if result.ready {
+				return result.version, nil
+			}
+		}
+
+		elapsed := time.Since(startTime).Round(time.Second)
+		_, _ = fmt.Fprintf(w, msgWaitingForCSV, elapsed)
```
den-rgb: double "Waiting for CSV" messages on each poll when CSV exists but isn't successful. tryCSVFromSubscription prints msgWaitingForCSVPhase and then waitForCSV prints msgWaitingForCSV

pkg/deps/install.go
```diff
@@ -0,0 +1,774 @@
...
+	// For "install all" mode, only install required deps unless --include-optional
+	isRequired := dep.Enabled == enabledTrue
+	if !isRequired && !c.IncludeOptional {
+		return false
+	}
+
+	return true
+}
+
+func (c *InstallCommand) isAlreadyInstalled(ctx context.Context, dep DependencyInfo) (bool, error) {
```
den-rgb, line 510: isAlreadyInstalled returns true without checking channel/source, so verifyExistingSubscription is unreachable on the install-all path.

pkg/deps/install.go
```diff
@@ -0,0 +1,774 @@
...
+		case <-ctx.Done():
+			return "", fmt.Errorf("context cancelled: %w", ctx.Err())
+		case <-ticker.C:
+		}
+	}
+}
+
+// matchesSubscription checks if a CSV name matches the subscription package.
+// CSV names follow the pattern "<package>.v<version>" (e.g., "kueue-operator.v0.10.0").
+func matchesSubscription(csvName, subName string) bool {
```
den-rgb: needs tests also, the strings.ReplaceAll(csvLower, "-", "") normalization is overly broad, use anchored prefix matching instead.

pkg/deps/install.go
```diff
@@ -0,0 +1,774 @@
...
+	csv, err := c.client.OLM().ClusterServiceVersions(namespace).Get(ctx, sub.Status.InstalledCSV, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return csvResult{ready: false}
+		}
+
+		return csvResult{err: fmt.Errorf("get CSV %s: %w", sub.Status.InstalledCSV, err)}
+	}
+
+	if csv.Status.Phase == operatorsv1alpha1.CSVPhaseSucceeded {
```
den-rgb: waitForCSV polls for the full timeout even onFailed, InstallCheckFailed phases, detect these

pkg/deps/install.go
```diff
@@ -0,0 +1,774 @@
...
+			continue
+		}
+
+		toInstall = append(toInstall, deps[depIdx])
+	}
+
+	return toInstall, nil
+}
+
+func (c *InstallCommand) shouldInstallDep(dep DependencyInfo) bool {
```
den-rgb, line 490: add test

pkg/deps/install.go
```diff
@@ -0,0 +1,774 @@
...
+	for _, d := range deps {
+		if d.Name == name {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (c *InstallCommand) runDryRun(_ context.Context, deps []DependencyInfo) error {
```
den-rgb, line 249: add test

pkg/deps/install.go
```diff
@@ -0,0 +1,774 @@
...
+	}
+
+	for _, dep := range toInstall {
+		c.printDryRunManifests(w, dep)
+	}
+
+	return nil
+}
+
+func (c *InstallCommand) printDryRunManifests(w io.Writer, dep DependencyInfo) {
```
den-rgb, line 269: np: YAML is missing managed-by labels that createNamespace/ensureOperatorGroup actually set.

cmd/deps/deps.go
```diff
@@ -12,36 +15,77 @@ import (
...
+	AddFlags(fs *pflag.FlagSet)
+	Complete() error
+	Validate() error
+	Run(ctx context.Context) error
+}
+
+// runCommand executes the Complete/Validate/Run lifecycle with error handling.
+//
+//nolint:wrapcheck // HandleError returns an already-handled error
+func runCommand(cmd *cobra.Command, c command, outputFormat string) error {
```
den-rgb: maybe use cmd.Command directly in runCommand to avoid the duplicate definition.

asanzgom:
@ShettyGaurav 
  ### Overall                                                                                                                                                                                 
                  
  This is a well-structured PR that extends the existing `deps` command with `install` and `graph` subcommands while preserving backward compatibility (bare `kubectl odh deps` still lists). 
  The code follows the project's established `Command` interface pattern and the `Complete/Validate/Run` lifecycle. The refactoring of `cmd/deps/deps.go` to extract a `runCommand` helper
  eliminates the repeated error-handling boilerplate nicely.                                                                                                                                  
                  
  ### Architecture / Design                                                                                                                                                                   
   
  **Positive:**                                                                                                                                                                               
  - Clean separation: `graph.go` and `install.go` are self-contained
  - The `command` interface in `cmd/deps/deps.go` + `runCommand` helper is a good abstraction for the three subcommands                                                                       
  - Concurrent subscription checks in `filterDepsToInstall` with bounded concurrency (`depInstallCheckConcurrency = 8`)                                                                       
  - `csvNamespaceListEveryNPolls` is a thoughtful optimization to reduce List calls                                                                                                           
  - Dry-run mode generates valid YAML manifests — useful for piping to `kubectl apply`                                                                                                        
  - `matchesSubscription` handles real-world OLM naming quirks (openshift- prefix, hyphen normalization)                                                                                      
                                                                                                                                                                                              
  **Concerns:**                                                                                                                                                                               
                                                                                                                                                                                              
  1. **Missing `managed-by` label on Subscription** (`install.go:518-534`). Namespace and OperatorGroup get `app.kubernetes.io/managed-by: odh-cli`, but the Subscription does not. This makes
   cleanup/identification inconsistent. If a future `uninstall` command relies on this label, it won't find Subscriptions.
                                                                                                                                                                                              
  2. **Dry-run YAML doesn't match actual resources** (`install.go:198-247`). The dry-run output omits the `managed-by` labels that `createNamespace` and `ensureOperatorGroup` actually add.  
  Users who diff dry-run vs actual will see discrepancies.
                                                                                                                                                                                              
  3. **Progress messages print after the action, not before** (`install.go:385-420`). `msgCreatingNamespace` prints *after* `createNamespace` returns. If `createNamespace` hangs or is slow, 
  the user sees nothing. Consider printing "Creating namespace..." before the call.
                                                                                                                                                                                              
  4. **`waitForCSV` doesn't fail fast on terminal CSV phases** (`install.go:599-636`). If the CSV enters `Failed` or `InstallCheckFailed`, the code keeps polling until timeout. This wastes 5
   minutes on an unrecoverable state. Check for `CSVPhaseFailed` and return immediately.
                                                                                                                                                                                              
  5. **`runInstall` discards per-dependency errors** (`install.go:258-270`). The function returns a sentinel `errInstallFailed` but drops the actual `result.Error` values. Consider using    
  `errors.Join` so callers get the full error chain.
                                                                                                                                                                                              
  6. **No timeout validation** (`install.go`). A user can pass `--timeout 0s` or `--timeout -1s`, which causes `waitForCSV` to immediately time out after one poll. `Validate()` should reject
   non-positive timeouts.
                                                                                                                                                                                              
  7. **Stale `TargetDep` on re-invocation** (`cmd/deps/deps.go:137-143`). The closure captures `installCommand` and only sets `TargetDep` when `len(args) > 0`. If the command is re-invoked  
  without args in-process (tests, embedded use), the previous value leaks. Reset it unconditionally:
     ```go                                                                                                                                                                                    
     installCommand.TargetDep = ""                                                                                                                                                            
     if len(args) > 0 {                                                                                                                                                                       
         installCommand.TargetDep = args[0]                                                                                                                                                   
     }                                                                                                                                                                                        
                                                                                                                                                                                              
  Fallback CSV version detection (check.go:103-106)                                                                                                                                           
                  
  The new findMatchingCSVVersion fallback is a reasonable improvement for cases where InstalledCSV is empty. However:                                                                         
  - It runs even when getVersionFromCSV returned an error (e.g., RBAC failure). The fallback should only trigger when err == nil && version == "".
  - findMatchingCSVVersion silently swallows List errors — RBAC/throttling failures become invisible. Consider returning (string, error).                                                     
                  
  matchesSubscription (install.go:700-723)                                                                                                                                                    
                  
  The hyphen-normalization logic is clever but could cause false positives. For example, "kueueoperatorx.v1.0.0" would match subscription "kueue-operator" because after removing hyphens,    
  "kueueoperatorx" has prefix "kueueoperator". The HasPrefix check doesn't verify a clean boundary. This is an edge case but worth unit-testing with adversarial inputs.
                                                                                                                                                                                              
  Missing tests: No unit tests for matchesSubscription — this is exactly the kind of tricky string logic that benefits from table-driven tests.                                               
  
  Test coverage                                                                                                                                                                               
                  
  - graph_test.go: Good coverage of lifecycle and output content
  - install_test.go: Only tests defaults, flags, and dry-run Complete. Missing:
    - matchesSubscription table-driven tests                                                                                                                                                  
    - Dry-run output verification (that YAML is valid)
    - shouldInstallDep behavior with various Enabled/TargetDep combinations                                                                                                                   
    - verifyExistingSubscription mismatch warnings
                                                                                                                                                                                              
  Minor           

  - graphHeaderWidth = 50 is hardcoded but the header length varies with version string. Cosmetic, but easy to compute dynamically.                                                           
  - shouldInstallDep silently allows installing disabled dependencies when explicitly targeted. Consider a warning on stderr: "Note: cert-manager is marked disabled in the manifest but
  installing as explicitly requested."                                                                                                                                                        
  - The install and graph subcommands pass "" as outputFormat to runCommand, so structured error output (--output json) won't work for these subcommands. Fine for now, but worth documenting.

pkg/deps/install.go
```diff
@@ -0,0 +1,922 @@
...
+	// If an OperatorGroup already exists, check if TargetNamespaces match
+	if len(list.Items) > 0 {
+		existing := &list.Items[0]
+		existingTargets, _, _ := unstructured.NestedStringSlice(existing.Object, "spec", "targetNamespaces")
+
+		if !slicesEqual(existingTargets, targetNamespaces) {
+			return false, fmt.Errorf(
+				"operatorgroup %s/%s exists with targetNamespaces %v, but requested %v",
+				namespace, existing.GetName(), existingTargets, targetNamespaces,
+			)
```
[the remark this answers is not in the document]
den-rgb: +1

pkg/deps/install.go
```diff
@@ -0,0 +1,930 @@
...
+	pollCount := 0
+
+	for {
+		if time.Now().After(deadline) {
+			return "", fmt.Errorf("timeout waiting for CSV after %v", c.Timeout)
+		}
+
+		pollCount++
+
+		result := c.tryCSVFromSubscription(ctx, w, namespace, subName, startTime)
```
den-rgb, line 871: wrap ctx with context.WithTimeout(ctx, c.Timeout) at the top of waitForCSV so the deadline propagates into all client calls.

pkg/deps/install.go
```diff
@@ -0,0 +1,937 @@
...
+
+	if !c.DryRun && !c.client.OLM().Available() {
+		return errors.New(msgOLMNotAvailableInst)
+	}
+
+	return nil
+}
+
+// Run executes the install command.
+func (c *InstallCommand) Run(ctx context.Context) error {
```
den-rgb, line 196: Run() hardcodes GetManifest(ctx, false), it has no --version flag, no --refresh. wire --version through or at minimum print which version is being used so that when a user who checks deps with kubectl odh deps --version 3.4.0 then runs kubectl odh deps install doesnt install deps for a different version silently

pkg/deps/install.go
```diff
@@ -0,0 +1,937 @@
...
+
+	if len(toInstall) == 0 {
+		_, _ = fmt.Fprintln(w, msgAllInstalled)
+
+		return nil
+	}
+
+	results := make([]InstallResult, 0, len(toInstall))
+
+	for _, dep := range toInstall {
```
den-rgb: each dep takes 5min to install creating a 25min install time for 5 deps, the check phase already parallelizes via `errgroup`. consider the same pattern for the CSV wait phase

pkg/deps/install.go
```diff
@@ -0,0 +1,937 @@
...
+		return true, nil
+	}
+
+	// No subscription - check for orphaned CSV (installed but subscription deleted)
+	return c.hasSucceededCSV(ctx, dep.Namespace, dep.Subscription)
+}
+
+func (c *InstallCommand) hasSucceededCSV(ctx context.Context, namespace, subName string) (bool, error) {
+	csvList, err := c.client.OLM().ClusterServiceVersions(namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
```
den-rgb, line 529: apierrors.IsNotFound(err) on a List call in hasSucceededCSV won't fire

den-rgb: maybe add meta.IsNoMatchError(err) if the intent is to tolerate a missing OLM CRD

pkg/deps/manifest.go
```diff
@@ -57,56 +74,131 @@ type Dependency struct {
...
-	Source    string `yaml:"source"`    // Catalog source (optional, defaults to redhat-operators)
+	Channel          string   `yaml:"channel"`
+	Name             string   `yaml:"name"`             // Subscription name
+	Namespace        string   `yaml:"namespace"`        // Operator namespace
+	Source           string   `yaml:"source"`           // Catalog source (optional, defaults to redhat-operators)
+	TargetNamespaces []string `yaml:"targetNamespaces"` // OperatorGroup target namespaces
+}
+
+// ConfigSpec contains post-install CR configuration.
+type ConfigSpec struct {
```
den-rgb: ConfigSpec is wired through parsing and display but never consumed. either drop it from this PR or add a // TODO: comment documenting the planned follow-up so it's not forgotten.

pkg/deps/install.go
den-rgb: also add refresh

pkg/deps/install.go
```diff
@@ -0,0 +1,937 @@
...
+		return nil, fmt.Errorf("convert to unstructured: %w", err)
+	}
+
+	return &unstructured.Unstructured{Object: data}, nil
+}
+
+// slicesEqualAsSet reports whether two string slices contain the same elements
+// regardless of order. Duplicates are handled via a count map so
+// ["a","a","b"] ≠ ["a","b"] as expected.
+func slicesEqualAsSet(a, b []string) bool {
```
den-rgb: np: slicesEqualAsSet performs multiset equality (counts matter), not set equality. slicesEqualUnordered would be more accurate

pkg/deps/install.go
ShettyGaurav: For Now The command is only valid for version 3.4  as we have only dependency graph for only 3.4 in odh-gitops.
 I will add --version and --refresh so that If dependencies graph of newer versions get added it should be compatible

pkg/deps/install.go
```diff
@@ -0,0 +1,1011 @@
...
+		if reason == "" {
+			reason = "unknown"
+		}
+
+		return csvResult{err: fmt.Errorf("CSV %s failed: %s", csv.Name, reason)}
+
+	default:
+		// Pending, Installing, InstallReady, Unknown, Replacing, Deleting - keep polling
+		elapsed := time.Since(startTime).Round(time.Second)
+		_, _ = fmt.Fprintf(w, msgWaitingForCSVPhase, csv.Status.Phase, elapsed)
```
den-rgb, line 830: tryCSVFromSubscription and waitForCSV write progress messages to w without holding the mutex from waitForCSVsParallel. multiple goroutines writing concurrently is a data race, wrap w in a synchronized writer or push all writes through the mutex.

pkg/deps/install.go
```diff
@@ -0,0 +1,1011 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to determine manifest version: %w", err)
+	}
+
+	c.manifestVersion = manifestVer
+
+	if c.Version != "" && !majorMinorMatch(c.Version, c.manifestVersion) {
+		return utilserrors.NewValidationError(
+			"VERSION_UNAVAILABLE",
+			fmt.Sprintf("dependency graph for version %q is not available; only version %s is supported", c.Version, c.manifestVersion),
```
den-rgb: change `dependancy graph` to `dependencies for version %q`

pkg/deps/install.go
```diff
@@ -0,0 +1,1011 @@
...
+	}
+
+	unstructuredOG, err := toUnstructured(og)
+	if err != nil {
+		return false, fmt.Errorf("create operatorgroup: %w", err)
+	}
+
+	_, err = c.client.Dynamic().Resource(gvr).Namespace(namespace).
+		Create(ctx, unstructuredOG, metav1.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
```
den-rgb: List returns no OGs, but another actor creates one before Create. the AlreadyExists branch returns (true, nil) this should be (false, nil) to match the "already exists" semantics.

ShettyGaurav:
@coderabbitai review

den-rgb (approved):
/lgtm

PR #74, opened 2026-05-08
feat: add events command for listing ODH events

ShettyGaurav:
## Description

This PR introduces the `odh events` command that provides visibility into Kubernetes events related to Open Data Hub / OpenShift AI resources.

**Key implementation details:**
- Uses `clusterhealth.RunRecentEvents()` for event collection
- Auto-discovers ODH namespaces from DSCInitialization CR
- Concurrent operator namespace discovery (checks multiple locations in parallel)
- Proper validation errors with actionable suggestions

##### Jira link :- https://redhat.atlassian.net/browse/RHOAIENG-54646

**Usage examples:**
```bash
kubectl odh events                    # Recent events (default: 5m)
kubectl odh events --since 1h         # Last hour
kubectl odh events --type Warning     # Warnings only
kubectl odh events -A -o json         # All ODH namespaces, JSON output
```

## Testing

## Review checklist

## Follow-up

A follow-up PR will add:
- `--component`: filter events by component (e.g., kserve, dashboard, workbenches)
- `--follow` / `-f`: stream new events as they occur

## Summary by CodeRabbit

* **New Features**
  * Added `events` subcommand to list and filter cluster events.
  * Supports filtering by event type (Warning/Normal), time range (up to 24 hours), and namespaces.
  * Output formats: table, JSON, and YAML.
  * Events sorted by recency with relative timestamps.

### DIFF
--- a/cmd/events/events.go
+++ b/cmd/events/events.go
@@ -1,0 +1,95 @@
+package events
+
+import (
+	"io"
+
+	"github.com/spf13/cobra"
+
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+
+	eventspkg "github.com/opendatahub-io/odh-cli/pkg/events"
+	clierrors "github.com/opendatahub-io/odh-cli/pkg/util/errors"
+)
+
+// handleErr writes the error in structured or text format and returns an already-handled error.
+//
+//nolint:wrapcheck // NewAlreadyHandledError is a sentinel, not meant to be wrapped
+func handleErr(w io.Writer, err error, outputFormat string) error {
+	if clierrors.WriteStructuredError(w, err, outputFormat) {
+		return clierrors.NewAlreadyHandledError(err)
+	}
+
+	clierrors.WriteTextError(w, err)
+
+	return clierrors.NewAlreadyHandledError(err)
+}
+
+const (
+	cmdName  = "events"
+	cmdShort = "Show events for ODH resources"
+)
+
+const cmdLong = `
+Shows Kubernetes events related to Open Data Hub / OpenShift AI resources.
+
+Events are fetched from ODH namespaces (applications, operator, monitoring)
+using the clusterhealth library. The namespaces are auto-detected from the
+DSCInitialization resource.
+`
+
+const cmdExample = `
+  # Show recent events
+  kubectl odh events
+
+  # Show events from the last 30 minutes
+  kubectl odh events --since 30m
+
+  # Show only warnings
+  kubectl odh events --type Warning
+
+  # All ODH namespaces, YAML output
+  kubectl odh events -A -o yaml
+`
+
+// AddCommand adds the events command to the root command.
+func AddCommand(root *cobra.Command, flags *genericclioptions.ConfigFlags) {
+	streams := genericiooptions.IOStreams{
+		In:     root.InOrStdin(),
+		Out:    root.OutOrStdout(),
+		ErrOut: root.ErrOrStderr(),
+	}
+
+	command := eventspkg.NewCommand(streams, flags)
+
+	cmd := &cobra.Command{
+		Use:           cmdName,
+		Short:         cmdShort,
+		Long:          cmdLong,
+		Example:       cmdExample,
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			errOut := cmd.ErrOrStderr()
+			outputFormat := command.OutputFormat
+
+			if err := command.Complete(); err != nil {
+				return handleErr(errOut, err, outputFormat)
+			}
+
+			if err := command.Validate(); err != nil {
+				return handleErr(errOut, err, outputFormat)
+			}
+
+			if err := command.Run(cmd.Context()); err != nil {
+				return handleErr(errOut, err, outputFormat)
+			}
+
+			return nil
+		},
+	}
+
+	command.AddFlags(cmd.Flags())
+
+	root.AddCommand(cmd)
+}
--- a/pkg/events/command.go
+++ b/pkg/events/command.go
@@ -1,0 +1,223 @@
+package events
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/spf13/pflag"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+
+	"github.com/opendatahub-io/odh-cli/pkg/cmd"
+	"github.com/opendatahub-io/odh-cli/pkg/util/client"
+	clierrors "github.com/opendatahub-io/odh-cli/pkg/util/errors"
+	"github.com/opendatahub-io/odh-cli/pkg/util/iostreams"
+)
+
+var _ cmd.Command = (*Command)(nil)
+
+const (
+	outputFormatTable = "table"
+	outputFormatJSON  = "json"
+	outputFormatYAML  = "yaml"
+
+	namespaceDiscoveryTimeout = 15 * time.Second
+	defaultSinceDuration      = 5 * time.Minute
+	maxSinceDuration          = 24 * time.Hour
+)
+
+// Command contains the events command configuration.
+type Command struct {
+	IO          iostreams.Interface
+	ConfigFlags *genericclioptions.ConfigFlags
+	Client      client.Client
+
+	// Flags
+	EventType          string
+	Since              time.Duration
+	AllNamespaces      bool
+	OutputFormat       string
+	OperatorNSOverride string
+
+	// Resolved fields (populated during Complete)
+	Namespace         string
+	NamespaceExplicit bool // true if user explicitly passed -n flag
+	ApplicationsNS    string
+	OperatorNS        string
+	MonitoringNS      string
+
+	// Clusterhealth integration
+	crClient crclient.Client
+}
+
+// NewCommand creates a new events Command with defaults.
+func NewCommand(
+	streams genericiooptions.IOStreams,
+	configFlags *genericclioptions.ConfigFlags,
+) *Command {
+	return &Command{
+		IO:           iostreams.NewIOStreams(streams.In, streams.Out, streams.ErrOut),
+		ConfigFlags:  configFlags,
+		OutputFormat: outputFormatTable,
+		OperatorNS:   client.DefaultRHOAIOperatorNamespace,
+	}
+}
+
+// AddFlags registers command-specific flags.
+func (c *Command) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&c.EventType, "type", "", "Filter events by type (Warning or Normal)")
+	fs.DurationVar(&c.Since, "since", defaultSinceDuration, "Only show events newer than this duration (e.g., 5m, 1h, 30s)")
+	fs.BoolVarP(&c.AllNamespaces, "all-namespaces", "A", false, "List events across all ODH namespaces")
+	fs.StringVarP(&c.OutputFormat, "output", "o", outputFormatTable, "Output format: table, json, or yaml")
+	fs.StringVar(&c.OperatorNSOverride, "operator-namespace", "", "Override the operator namespace (auto-detected from OLM/CSV)")
+}
+
+// Complete resolves derived fields after flag parsing.
+func (c *Command) Complete() error {
+	restConfig, err := client.NewRESTConfig(c.ConfigFlags, client.DefaultQPS, client.DefaultBurst)
+	if err != nil {
+		return clierrors.ErrConfigFailed(err)
+	}
+
+	k8sClient, err := client.NewClientWithConfig(restConfig)
+	if err != nil {
+		return clierrors.ErrClientFailed(err)
+	}
+
+	c.Client = k8sClient
+
+	crClient, err := client.NewControllerRuntimeClient(restConfig)
+	if err != nil {
+		return clierrors.ErrCRClientFailed(err)
+	}
+
+	c.crClient = crClient
+
+	// Check if user explicitly passed -n flag (vs namespace from kubeconfig)
+	c.NamespaceExplicit = c.ConfigFlags.Namespace != nil && *c.ConfigFlags.Namespace != ""
+
+	if !c.AllNamespaces {
+		ns, _, err := c.ConfigFlags.ToRawKubeConfigLoader().Namespace()
+		if err != nil {
+			return clierrors.ErrNamespaceFailed(err)
+		}
+
+		c.Namespace = ns
+	}
+
+	return nil
+}
+
+// Validate checks that all options are valid before execution.
+func (c *Command) Validate() error {
+	if c.AllNamespaces && c.ConfigFlags.Namespace != nil && *c.ConfigFlags.Namespace != "" {
+		return clierrors.NewValidationError(
+			"INVALID_FLAGS",
+			"--all-namespaces and --namespace are mutually exclusive",
+			"Use either -A or -n <namespace>, not both",
+		)
+	}
+
+	switch c.OutputFormat {
+	case outputFormatTable, outputFormatJSON, outputFormatYAML:
+	default:
+		return clierrors.NewValidationError(
+			"INVALID_OUTPUT_FORMAT",
+			fmt.Sprintf("invalid output format %q", c.OutputFormat),
+			"Use one of: table, json, yaml",
+		)
+	}
+
+	if c.EventType != "" && c.EventType != "Warning" && c.EventType != "Normal" {
+		return clierrors.NewValidationError(
+			"INVALID_EVENT_TYPE",
+			fmt.Sprintf("invalid event type %q", c.EventType),
+			"Use --type Warning or --type Normal",
+		)
+	}
+
+	if c.Since < 0 {
+		return clierrors.NewValidationError(
+			"INVALID_DURATION",
+			"--since must be a non-negative duration",
+			"Use 0 or a positive value like --since 5m or --since 1h",
+		)
+	}
+
+	if c.Since > maxSinceDuration {
+		return clierrors.NewValidationError(
+			"INVALID_DURATION",
+			fmt.Sprintf("--since duration %v exceeds maximum of 24h", c.Since),
+			"Use --since with values like 5m, 1h, or 24h",
+		)
+	}
+
+	return nil
+}
+
+// Run executes the events command.
+func (c *Command) Run(ctx context.Context) error {
+	if err := c.discoverNamespaces(ctx); err != nil {
+		return err
+	}
+
+	return c.listEvents(ctx)
+}
+
+// discoverNamespaces resolves ODH namespaces using centralized discovery helpers.
+// Called from Run() so it respects the command context for cancellation.
+func (c *Command) discoverNamespaces(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, namespaceDiscoveryTimeout)
+	defer cancel()
+
+	// Use combined helper for single API call
+	namespaces, err := client.GetDSCINamespaces(ctx, c.Client)
+	if err != nil && !apierrors.IsNotFound(err) {
+		_, _ = fmt.Fprintf(c.IO.ErrOut(), "Warning: failed to get DSCI namespaces: %v\n", err)
+	}
+
+	c.ApplicationsNS = namespaces.Applications
+	c.MonitoringNS = namespaces.Monitoring
+
+	if c.OperatorNSOverride != "" {
+		c.OperatorNS = c.OperatorNSOverride
+
+		return nil
+	}
+
+	operatorNS, err := client.DiscoverOperatorNamespace(ctx, c.Client)
+	if err == nil {
+		c.OperatorNS = operatorNS
+
+		return nil
+	}
+
+	// Check if this is a "not found" error (operator not installed) vs RBAC/other errors
+	var structured *clierrors.StructuredError
+	if !errors.As(err, &structured) || structured.Category != clierrors.CategoryNotFound {
+		// Forbidden/RBAC errors should fail - user can use --operator-namespace
+		return fmt.Errorf("discovering operator namespace: %w", err)
+	}
+
+	_, _ = fmt.Fprint(c.IO.ErrOut(), "Warning: operator namespace not found, using default\n")
+	c.OperatorNS = client.DefaultRHOAIOperatorNamespace
+
+	return nil
+}
+
+// listEvents fetches and displays events.
+func (c *Command) listEvents(ctx context.Context) error {
+	events, err := c.fetchEvents(ctx)
+	if err != nil {
+		return err
+	}
+
+	sortEventsByTime(events)
+
+	return c.renderOutput(events)
+}
--- a/pkg/events/filter.go
+++ b/pkg/events/filter.go
@@ -1,0 +1,80 @@
+package events
+
+import (
+	"cmp"
+	"context"
+	"slices"
+	"time"
+
+	"github.com/opendatahub-io/opendatahub-operator/pkg/clusterhealth"
+
+	clierrors "github.com/opendatahub-io/odh-cli/pkg/util/errors"
+)
+
+const eventFetchTimeout = 30 * time.Second
+
+// fetchEvents retrieves events using the clusterhealth library.
+func (c *Command) fetchEvents(ctx context.Context) ([]clusterhealth.EventInfo, error) {
+	namespaces, err := c.getTargetNamespaces()
+	if err != nil {
+		return nil, err
+	}
+
+	fetchCtx, cancel := context.WithTimeout(ctx, eventFetchTimeout)
+	defer cancel()
+
+	cfg := clusterhealth.RecentEventsConfig{
+		Client:     c.crClient,
+		Namespaces: namespaces,
+		Since:      c.Since,
+		EventType:  c.EventType,
+	}
+
+	events, err := clusterhealth.RunRecentEvents(fetchCtx, cfg)
+	if err != nil {
+		return nil, clierrors.ErrEventsFetchFailed(err)
+	}
+
+	return events, nil
+}
+
+// getTargetNamespaces returns the namespaces to query for events.
+// For -n <namespace>, returns ONLY that namespace (exclusive scope like kubectl).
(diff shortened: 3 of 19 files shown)


ShettyGaurav:
@coderabbitai review

pkg/events/filter.go
```diff
@@ -0,0 +1,66 @@
...
+	add(c.ApplicationsNS)
+	add(c.OperatorNS)
+	add(c.MonitoringNS)
+
+	// Include current namespace context unless --all-namespaces
+	if !c.AllNamespaces && c.Namespace != "" {
+		add(c.Namespace)
+	}
+
+	if len(namespaces) == 0 {
```
den-rgb, line 67: falling back to "default" when no ODH namespaces are discovered silently returns unrelated events or a misleading "No events found". return an error instead

pkg/events/command.go
```diff
@@ -0,0 +1,243 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), namespaceDiscoveryTimeout)
+	defer cancel()
+
+	dsci, err := client.GetDSCInitialization(ctx, c.Client)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("getting DSCInitialization: %w", err)
+	}
+
+	if dsci != nil {
+		c.ApplicationsNS, _ = jq.Query[string](dsci, ".spec.applicationsNamespace")
```
den-rgb: if the DSCI schema changes these silently fall into the "default" namespace fallback. handle errors like pkg/status/namespace.go:82 and pkg/util/client/helpers.go:368 do

pkg/events/command.go
```diff
@@ -0,0 +1,243 @@
...
+	if !c.AllNamespaces {
+		ns, _, err := c.ConfigFlags.ToRawKubeConfigLoader().Namespace()
+		if err != nil {
+			return fmt.Errorf("determining namespace: %w", err)
+		}
+
+		c.Namespace = ns
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), namespaceDiscoveryTimeout)
```
den-rgb: context.Background() here ignores cmd.Context() so Ctrl+C won't cancel namespace discovery for up to 10s. move the DSCI lookup and discoverOperatorNS into Run() where the proper context is available, similar to how status does it

pkg/events/output.go
```diff
@@ -0,0 +1,141 @@
...
+
+	tw := tabwriter.NewWriter(w, 0, 0, tabwriterPadding, ' ', 0)
+
+	if _, err := fmt.Fprint(tw, tableHeader); err != nil {
+		return fmt.Errorf("writing header: %w", err)
+	}
+
+	for _, e := range events {
+		age := formatAge(e.LastTime)
+		object := fmt.Sprintf("%s/%s", e.Kind, e.Name)
```
den-rgb: EventInfo has Namespace and Count fields that aren't displayed. at minimum show the namespace column when -A is used, like kubectl get events -A does

cmd/events/events.go
```diff
@@ -0,0 +1,111 @@
...
+
+	return clierrors.NewAlreadyHandledError(err)
+}
+
+const (
+	cmdName  = "events"
+	cmdShort = "Show events for ODH resources"
+)
+
+const cmdLong = `
```
den-rgb, line 33: np: cmdLong embeds examples that overlap with cmdExample, cobra renders both, so users see them twice. remove examples from cmdLong

den-rgb: or first verify that users dont see them twice, and if they do remove it

pkg/events/command.go
```diff
@@ -0,0 +1,265 @@
...
+
+			go func(namespace, deployName string) {
+				defer wg.Done()
+
+				if subCtx.Err() != nil {
+					return
+				}
+
+				_, err := r.GetResource(subCtx, resources.Deployment, deployName, client.InNamespace(namespace))
+				if err != nil {
```
den-rgb: GetResource errors are dropped unconditionally here, RBAC Forbidden and API timeouts take the same "not found" path. distinguish IsNotFound from other errors

pkg/events/output.go
den-rgb: i'd still consider adding a count field to distinguish once off errors from reoccuring ones

pkg/events/filter.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, ErrNoNamespacesDiscovered
+	}
+
+	return namespaces, nil
+}
+
+// sortEventsByTime sorts events in place by timestamp, most recent first.
+func sortEventsByTime(events []clusterhealth.EventInfo) []clusterhealth.EventInfo {
```
den-rgb: sort.Slice mutates in place but the function also returns the slice, making it look like a copy. either drop the return or copy before sorting. slices.SortFunc would also work here

pkg/events/command.go
```diff
@@ -0,0 +1,284 @@
...
+
+	sorted := sortEventsByTime(events)
+
+	return c.renderOutput(sorted)
+}
+
+// discoverOperatorNS finds the operator namespace by checking well-known locations.
+// Deployment lookups run concurrently and cancel remaining work after the first hit.
+// Returns the discovered namespace and any non-NotFound errors encountered.
+func discoverOperatorNS(ctx context.Context, r client.Reader) (string, error) {
```
den-rgb: discoverOperatorNS reimplements pkg/status/namespace.go:discoverOperatorNamespaceFromDefaults with added concurrency that provides no measurable benefit for 6 sequential API calls at CLI startup. it also misses the OLM-based discovery path that status uses. extract the status package's discovery into a shared helper and reuse it here

pkg/events/command.go
den-rgb: implement more error checking here similar pattern to how status returns ErrOperatorNamespaceNotFound(), its missing evaluation for errors like 'Forbidden'

pkg/events/filter.go
```diff
@@ -0,0 +1,77 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+
+	"github.com/opendatahub-io/opendatahub-operator/pkg/clusterhealth"
+)
+
+// ErrNoNamespacesDiscovered is returned when no ODH namespaces could be found.
+var ErrNoNamespacesDiscovered = errors.New("no ODH namespaces discovered; ensure ODH/RHOAI is installed")
```
den-rgb: ErrNoNamespacesDiscovered is a plain error but the CLI renders structured errors in JSON/YAML mode. make this a *clierrors.StructuredError consistent with pkg/status/errors.go:ErrNoDSCIFound

pkg/events/output_test.go
```diff
@@ -0,0 +1,263 @@
...
+		{"3 days ago", now.Add(-3 * 24 * time.Hour), "3d"},
+		{"future time", now.Add(1 * time.Hour), "0s"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := formatAge(tt.time)
+			if result != tt.expected {
+				t.Errorf("formatAge() = %q, expected %q", result, tt.expected)
+			}
```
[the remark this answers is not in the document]
den-rgb: +1

den-rgb:
no unit test covers discoverOperatorNS despite it being the most complex function in the PR. add cases for first-hit-cancels, all-not-found, and Forbidden error

pkg/events/filter.go
ShettyGaurav: Addressed.

pkg/events/output.go
ShettyGaurav: Addressed

pkg/events/filter.go
ShettyGaurav: Addressed

ShettyGaurav:
@coderabbitai review

ShettyGaurav:
> no unit test covers discoverOperatorNS despite it being the most complex function in the PR. add cases for first-hit-cancels, all-not-found, and Forbidden error

Centralized namespace discovery and Added Tests for it

pkg/events/filter.go
ShettyGaurav: Addressed

pkg/events/command.go
```diff
@@ -0,0 +1,206 @@
...
+	return c.listEvents(ctx)
+}
+
+// discoverNamespaces resolves ODH namespaces using centralized discovery helpers.
+// Called from Run() so it respects the command context for cancellation.
+func (c *Command) discoverNamespaces(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, namespaceDiscoveryTimeout)
+	defer cancel()
+
+	appNS, err := client.GetApplicationsNamespace(ctx, c.Client)
```
den-rgb: GetApplicationsNamespace and GetMonitoringNamespace each call GetDSCInitialization independently, combine them similar to how pkg/status/namespace.go:discoverNamespaces does it

pkg/util/errors/common.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// ErrOperatorNamespaceNotFound creates a structured error when operator namespace cannot be discovered.
+func ErrOperatorNamespaceNotFound() *StructuredError {
+	return &StructuredError{
+		Code:       CodeNotFound,
+		Message:    "could not discover operator namespace",
+		Category:   CategoryNotFound,
+		Retriable:  false,
+		Suggestion: "Use --operator-namespace to specify the operator namespace",
```
den-rgb, line 91: verify the events command has --operator-namespace , since this shared error suggests it. either make the suggestion generic or accept it as a parameter

pkg/events/command.go
```diff
@@ -0,0 +1,206 @@
...
+
+	monNS, err := client.GetMonitoringNamespace(ctx, c.Client)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return clierrors.ErrDSCIFailed(err)
+	}
+
+	c.MonitoringNS = monNS
+
+	operatorNS, err := client.DiscoverOperatorNamespace(ctx, c.Client)
+	if err != nil {
```
den-rgb: silently falling back to redhat-ods-operator on Forbidden hides RBAC issues. consider distinguishing Forbidden from not-found here

pkg/util/errors/common.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// ErrDSCIFailed creates a structured error for DSCInitialization retrieval failures.
+func ErrDSCIFailed(err error) *StructuredError {
+	return classifyOrWrap(err, CodeDSCIFailed, "failed to get DSCInitialization",
+		SuggestCheckODH, CategoryInternal, true)
+}
+
+// ErrDSCIQueryFailed creates a structured error for DSCI field query failures.
+func ErrDSCIQueryFailed(field string, err error) *StructuredError {
```
den-rgb: unused?

pkg/events/output.go
```diff
@@ -0,0 +1,158 @@
...
+		return clierrors.ErrRenderFailed("YAML", err)
+	}
+
+	return nil
+}
+
+// toEventOutputList converts EventInfo slice to a structure suitable for JSON/YAML rendering.
+// Always returns consistent EventList shape for stable machine consumption.
+func toEventOutputList(events []clusterhealth.EventInfo) any {
+	return map[string]any{
```
den-rgb, line 188: np: consider adding "apiVersion": "v1" to match the Kubernetes convention for list-kind output shapes

pkg/util/client/discovery.go
```diff
@@ -0,0 +1,135 @@
...
+	if err == nil {
+		return ns, nil
+	}
+
+	// Fall back to OLM-based discovery (slower but comprehensive)
+	if info := DiscoverOperatorFromOLM(ctx, c); info != nil && info.Namespace != "" {
+		return info.Namespace, nil
+	}
+
+	return "", clierrors.ErrOperatorNamespaceNotFound()
```
den-rgb: DiscoverOperatorNamespace drops the error from discoverOperatorNamespaceFromDefaults and returns ErrOperatorNamespaceNotFound when OLM also fails. if defaults returned Forbidden, the caller in command.go:207 sees CategoryNotFound and silently falls back. return defaultsErr instead of a fresh ErrOperatorNamespaceNotFound() so Forbidden propagates to the user.

pkg/events/command.go
```diff
@@ -0,0 +1,228 @@
...
+	defer cancel()
+
+	// Fetch DSCI once, extract both namespaces
+	dsci, err := client.GetDSCInitialization(ctx, c.Client)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return clierrors.ErrDSCIFailed(err)
+	}
+
+	if dsci != nil {
+		c.ApplicationsNS, _ = jq.Query[string](dsci, ".spec.applicationsNamespace")
```
den-rgb: q.Query errors are discarded with _ here and on line 184. if the DSCI schema changes, both namespaces silently become empty. check for errors that aren't jq.ErrNotFound and log a warning, similar to how GetApplicationsNamespace in helpers.go handles it.

pkg/events/output_test.go
```diff
@@ -0,0 +1,294 @@
...
+	"testing"
+	"time"
+
+	"github.com/opendatahub-io/opendatahub-operator/pkg/clusterhealth"
+)
+
+func TestFormatAge(t *testing.T) {
+	// Use values well away from boundaries to avoid clock-drift flakes.
+	// e.g., 25s won't flip to "1m" even with a few seconds of drift.
+	now := time.Now()
```
den-rgb: formatAge calls time.Since() internally while the test captures now separately. the "25s" case will flake under scheduler pressure. inject a reference time into formatAge or use a tolerance set in assertions. CodeRabbit flagged this in all 3 review rounds.

pkg/util/client/helpers.go
```diff
@@ -387,6 +387,33 @@ func GetApplicationsNamespace(ctx context.Context, r Reader) (string, error) {
 	return namespace, nil
 }
 
+// GetMonitoringNamespace retrieves the monitoring namespace from DSCInitialization.
+// Returns the namespace string if found in spec.monitoring.namespace.
+// If not set or empty, falls back to the applications namespace.
+// Returns empty string and error only for non-recoverable failures.
+func GetMonitoringNamespace(ctx context.Context, r Reader) (string, error) {
```
den-rgb: where is this meant to be used? i see tests but no usage, maybe use it in discoverNamespaces

pkg/events/command.go
```diff
@@ -0,0 +1,228 @@
...
+)
+
+var _ cmd.Command = (*Command)(nil)
+
+const (
+	outputFormatTable = "table"
+	outputFormatJSON  = "json"
+	outputFormatYAML  = "yaml"
+
+	namespaceDiscoveryTimeout = 60 * time.Second
```
den-rgb: is this too long?

pkg/events/output.go
ShettyGaurav: Addressed

pkg/util/errors/common.go
ShettyGaurav: Addressed

pkg/events/command.go
ShettyGaurav: Adressed

pkg/util/errors/common.go
ShettyGaurav: Addressed

pkg/events/command.go
ShettyGaurav: Addressed

pkg/events/command.go
ShettyGaurav: Resolved

pkg/util/client/discovery.go
ShettyGaurav: Addressed - now returns defaultsErr instead of creating a new ErrOperatorNamespaceNotFound(). Forbidden/RBAC errors from discoverOperatorNamespaceFromDefaults now propagate to the caller.

pkg/events/command.go
ShettyGaurav: Addressed - now checks for errors that aren't jq.ErrNotFound and logs a warning to stderr for both applicationsNamespace and monitoring.namespace queries.

pkg/events/command.go
ShettyGaurav: Addressed : Reduced to 15 Seconds

pkg/events/output_test.go
ShettyGaurav: Addressed - added formatAgeFrom(t, now) with injected reference time. Test uses fixed clock.

pkg/util/client/helpers.go
ShettyGaurav: Addressed - removed unused function and its tests. Events command handles monitoring namespace inline.

pkg/events/command.go
ShettyGaurav: Addressed

pkg/events/command.go
ShettyGaurav: Addressed - moved to Run() and uses context.WithTimeout(ctx, ...) wrapping cmd context. Ctrl+C cancellation works.

pkg/events/command.go
ShettyGaurav: Addressed - now checks for errors that aren't jq.ErrNotFound and logs warnings to stderr.

cmd/events/events.go
ShettyGaurav: Resolved - cmdLong contains only description, examples are in cmdExample only. No duplication.

ShettyGaurav:
@coderabbitai review

pkg/events/output.go
```diff
@@ -0,0 +1,163 @@
...
+		object := fmt.Sprintf("%s/%s", e.Kind, e.Name)
+		message := truncateMessage(e.Message)
+
+		if showNamespace {
+			if _, err := fmt.Fprintf(tw, "%s\t%s\t%s\t%s\t%s\t%d\t%s\n", e.Namespace, age, e.Type, e.Reason, object, e.Count, message); err != nil {
+				return clierrors.ErrRenderFailed("table", err)
+			}
+		} else {
+			if _, err := fmt.Fprintf(tw, "%s\t%s\t%s\t%s\t%d\t%s\n", age, e.Type, e.Reason, object, e.Count, message); err != nil {
+				return clierrors.ErrRenderFailed("table", err)
```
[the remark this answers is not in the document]
ShettyGaurav: Addressed - added sanitizeForTerminal() to strip control characters from event messages before table rendering. Includes tests

pkg/util/errors/common.go
```diff
@@ -0,0 +1,108 @@
...
+	SuggestCheckCluster    = "Ensure the cluster is running and accessible"
+	SuggestCheckODH        = "Ensure ODH/RHOAI is properly installed"
+	SuggestRetry           = "Try again; if the problem persists, check cluster connectivity"
+)
+
+// classifyOrWrap attempts to classify the error. If classification succeeds,
+// it prepends the context message. Otherwise, it creates a new StructuredError.
+func classifyOrWrap(err error, code, context, suggestion string, category ErrorCategory, retriable bool) *StructuredError {
+	if classified := Classify(err); classified != nil {
+		classified.Message = fmt.Sprintf("%s: %s", context, classified.Message)
```
den-rgb: copy *classified before mutating .Message. Classify() returns a pointer into the error chain, so in-place mutation corrupts the original. wrapped := *classified; wrapped.Message = ...; return &wrapped

pkg/events/filter.go
```diff
@@ -0,0 +1,72 @@
...
+	if len(namespaces) == 0 {
+		return nil, clierrors.ErrNoNamespacesDiscovered()
+	}
+
+	return namespaces, nil
+}
+
+// sortEventsByTime sorts events in place by timestamp, most recent first.
+func sortEventsByTime(events []clusterhealth.EventInfo) {
+	slices.SortFunc(events, func(a, b clusterhealth.EventInfo) int {
```
den-rgb: slices.SortFunc is unstable. events with identical LastTime shuffle between runs. use slices.SortStableFunc for deterministic output

pkg/events/output.go
```diff
@@ -0,0 +1,177 @@
...
+	} else {
+		if _, err := fmt.Fprint(tw, tableHeader); err != nil {
+			return clierrors.ErrRenderFailed("table", err)
+		}
+	}
+
+	for _, e := range events {
+		age := formatAge(e.LastTime)
+		object := fmt.Sprintf("%s/%s", e.Kind, e.Name)
+		message := truncateMessage(sanitizeForTerminal(e.Message))
```
den-rgb, line 133: np: e.Reason is also untrusted controller-set text. consider sanitizeForTerminal(e.Reason) for consistency with the message sanitization

pkg/status/namespace_test.go
den-rgb: these mocks return errors.New("not found") which doesn't satisfy apierrors.IsNotFound(), maybe use apierrors.NewNotFound(schema.GroupResource{Resource: "deployments"}, name) so the tests exercise the actual NotFound-skip path in discoverOperatorNamespaceFromDefaults

pkg/util/errors/common.go
ShettyGaurav: Fixed. Now copies the struct before mutating

pkg/events/filter.go
ShettyGaurav: Addressed Now uses slices.SortStableFunc

pkg/events/output.go
ShettyGaurav: Addressed Added reason := sanitizeForTerminal(e.Reason) for consistency with message sanitization.

pkg/status/namespace_test.go
ShettyGaurav: Addressed

pkg/events/filter.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	cfg := clusterhealth.RecentEventsConfig{
+		Client:     c.crClient,
+		Namespaces: namespaces,
+		Since:      c.Since,
+		EventType:  c.EventType,
+	}
+
+	events, err := clusterhealth.RunRecentEvents(ctx, cfg)
```
den-rgb: RunRecentEvents receives the raw context with no timeout. discoverNamespaces gets 15s but the heavier event fetch is unbounded. wrap with context.WithTimeout or share a budget from Run()

pkg/events/command.go
```diff
@@ -0,0 +1,237 @@
...
+	if err := c.discoverNamespaces(ctx); err != nil {
+		return err
+	}
+
+	return c.listEvents(ctx)
+}
+
+// discoverNamespaces resolves ODH namespaces using centralized discovery helpers.
+// Called from Run() so it respects the command context for cancellation.
+func (c *Command) discoverNamespaces(ctx context.Context) error {
```
den-rgb, line 174: discoverNamespaces orchestrates DSCI fetch, jq extraction, operator discovery, and the CategoryNotFound vs RBAC branching at line 216, none of it is tested. add mock-based cases for at least DSCI-not-found fallback and RBAC error propagation

pkg/events/output.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+	} else {
+		if _, err := fmt.Fprint(tw, tableHeader); err != nil {
+			return clierrors.ErrRenderFailed("table", err)
+		}
+	}
+
+	for _, e := range events {
+		age := formatAge(e.LastTime)
+		object := fmt.Sprintf("%s/%s", e.Kind, e.Name)
```
den-rgb: np: e.Type, e.Kind, e.Name, e.Namespace are printed unsanitized while e.Reason and e.Message are sanitized. for consistency, sanitize all rendered fields

pkg/events/output.go
```diff
@@ -0,0 +1,178 @@
...
+	if len(r) <= maxMessageLen {
+		return msg
+	}
+
+	return string(r[:maxMessageLen-3]) + "..."
+}
+
+// sanitizeForTerminal removes control characters from untrusted input
+// to prevent terminal escape sequence injection (CWE-150).
+func sanitizeForTerminal(s string) string {
```
den-rgb, line 74: np: stripping only \x1b might leave bracket remnants like [31mred[0m. consider stripping full ANSI sequences with a regex or strings.Map that also drops [ when preceded by a stripped ESC, or at least verify this doesnt happen

pkg/events/filter.go
ShettyGaurav: Added 30s timeout in `fetchEvents()`

pkg/events/command.go
ShettyGaurav: Added 7 test cases in pkg/events/discover_test.go

pkg/events/output.go
ShettyGaurav: All rendered fields now pass through sanitizeForTerminal()

pkg/events/output.go
ShettyGaurav: Added regex to strip full ANSI CSI sequences before control char stripping

pkg/events/filter.go
```diff
@@ -0,0 +1,79 @@
...
+	return events, nil
+}
+
+// getTargetNamespaces returns the namespaces to query for events.
+// For -n <namespace>, returns ONLY that namespace (exclusive scope like kubectl).
+// For --all-namespaces or no flags, returns ODH namespaces (apps, operator, monitoring).
+// Returns ErrNoNamespacesDiscovered if no namespaces could be determined.
+func (c *Command) getTargetNamespaces() ([]string, error) {
+	// If user specified -n <namespace>, return ONLY that namespace (exclusive)
+	if !c.AllNamespaces && c.Namespace != "" {
```
den-rgb: the default kubectl odh events invocation (no flags) will resolve c.Namespace from kubeconfig in Complete(), so it hits the exclusive -n path and returns only the kubeconfig namespace, not the ODH namespaces. is this intentional?

pkg/events/command.go
```diff
@@ -0,0 +1,228 @@
...
+	ctx, cancel := context.WithTimeout(ctx, namespaceDiscoveryTimeout)
+	defer cancel()
+
+	// Use shared helpers for namespace discovery
+	appNS, err := client.GetApplicationsNamespace(ctx, c.Client)
```
den-rgb: np: GetApplicationsNamespace and GetMonitoringNamespace each call GetDSCInitialization independently, two API calls instead of one. fine for CLI startup but could be a single-fetch helper

pkg/events/filter.go
ShettyGaurav: Fixed by adding a NamespaceExplicit field that tracks whether the user explicitly passed -n vs the namespace coming from kubeconfig.

pkg/events/command.go
ShettyGaurav: Added GetDSCINamespaces() helper that fetches both namespaces in a single API call
and also removed GetMonitoringNamespace()

den-rgb (approved):
/lgtm

PR #79, opened 2026-05-19
RHOAIENG-61428: Wire migrate command framework into rhai-cli

asmigala:
Register the migrate command (list, prepare, run) in the CLI entry point and add lifecycle phase support for migration actions.

- Wire migrate command into cmd/main.go
- Add ActionPhase type (pre-upgrade, post-upgrade, pre-enablement) with Phase() method on the Action interface
- Add --phase flag to list, run, and prepare subcommands
- Auto-detect phase from cluster version comparison when --phase is omitted
- Add phase column to migrate list table output
- Add phase filtering to ActionRegistry and Executor
- Add shell completion for --phase and --migration flags
- Add DataScienceClusterV1 resource type for 2.x cluster compat
- Return non-zero exit when no migrations match
- Report skipped steps in final output when user declines prompts
- Show phase auto-detection hint in list output

## Testing

## Review checklist

## Summary by CodeRabbit

* **New Features**
  * Added top-level migrate subcommand and full --phase support (pre-upgrade, post-upgrade, pre-enablement) with auto-detection, phase-aware list/run/prepare behavior, new PHASE column, improved help/examples, and shell tab-completion for phase and migration IDs.
  * Added phase-aware action lifecycle APIs, detection behavior, skipped-steps reporting, and improved backup/no-backup output.
  * Improved Kueue-related cluster resource handling and added v1 resource type.

* **Tests**
  * Broadened unit test coverage across phase detection, action registry, executor, results, and migrate commands.

### DIFF
--- a/pkg/migrate/action/executor.go
+++ b/pkg/migrate/action/executor.go
@@ -37,8 +37,9 @@ func (e *Executor) ExecuteSelective(
 	target Target,
 	pattern string,
 	group ActionGroup,
+	phase ActionPhase,
 ) ([]ActionExecution, error) {
-	actions, err := e.registry.ListByPattern(pattern, group)
+	actions, err := e.registry.ListByFilter(pattern, group, phase)
 	if err != nil {
 		return nil, fmt.Errorf("selecting actions: %w", err)
 	}
--- a/pkg/migrate/command_prepare.go
+++ b/pkg/migrate/command_prepare.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"time"
 
@@ -28,8 +29,10 @@ type PrepareCommand struct {
 	OutputDir     string
 	MigrationIDs  []string
 	TargetVersion string
+	Phase         string
 
 	parsedTargetVersion *semver.Version
+	parsedPhase         action.ActionPhase
 
 	// registry is the action registry for this command instance.
 	// Explicitly populated to avoid global state and enable test isolation.
@@ -49,6 +52,10 @@ func NewPrepareCommand(streams genericiooptions.IOStreams) *PrepareCommand {
 	}
 }
 
+func (c *PrepareCommand) ActionIDs() []string {
+	return c.registry.ActionIDs()
+}
+
 func (c *PrepareCommand) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVarP(&c.Verbose, "verbose", "v", false, flagDescPrepareVerbose)
 	fs.DurationVar(&c.Timeout, "timeout", c.Timeout, flagDescPrepareTimeout)
@@ -57,6 +64,7 @@ func (c *PrepareCommand) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.OutputDir, "output-dir", "", flagDescPrepareOutputDir)
 	fs.StringArrayVarP(&c.MigrationIDs, "migration", "m", []string{}, flagDescPrepareMigration)
 	fs.StringVar(&c.TargetVersion, "target-version", "", flagDescPrepareTargetVersion)
+	fs.StringVar(&c.Phase, "phase", "", flagDescPreparePhase)
 
 	// Throttling settings
 	fs.Float32Var(&c.QPS, "qps", c.QPS, "Kubernetes API QPS limit (queries per second)")
@@ -70,6 +78,11 @@ func (c *PrepareCommand) Complete() error {
 
 	// Always enable verbose for migrate prepare
 	c.Verbose = true
+
+	c.parsedPhase = ""
+	if c.Phase != "" {
+		c.parsedPhase = action.ActionPhase(c.Phase)
+	}
 
 	if c.TargetVersion != "" {
 		// Use ParseTolerant to accept partial versions (e.g., "3.0" → "3.0.0")
@@ -94,12 +107,16 @@ func (c *PrepareCommand) Validate() error {
 		return fmt.Errorf("validating shared options: %w", err)
 	}
 
-	if len(c.MigrationIDs) == 0 {
-		return errors.New("--migration flag is required")
+	if len(c.MigrationIDs) == 0 && c.Phase == "" {
+		return errors.New("--migration flag is required (or use --phase to prepare all actions for a lifecycle phase)")
 	}
 
 	if c.TargetVersion == "" {
 		return errors.New("--target-version flag is required")
+	}
+
+	if err := c.parsedPhase.Validate(); err != nil {
+		return fmt.Errorf("validating phase: %w", err)
 	}
 
 	return nil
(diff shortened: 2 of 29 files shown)


pkg/migrate/phase.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// resolvePhaseAndMigrations determines the effective phase and populates migration IDs
+// when they aren't explicitly provided. Returns the effective phase and the migration IDs.
+//
+// FIXME(team): When --migration is provided without --phase, the auto-detected phase still
+// applies and the execution loop skips actions whose phase doesn't match. This means
+// explicitly-requested migrations can be silently skipped (e.g., running a pre-upgrade
+// action on a cluster already at the target version). Discuss whether explicit --migration
+// should bypass phase filtering entirely.
```
chambridge: 👍  on flagging this. 

**My vote:** explicit `--migration` should bypass phase filtering entirely.
The user asked for it by name. If you want a guardrail, error > silent skip.

Thoughts @lburgazzoli ?

pkg/migrate/command_run.go
```diff
@@ -104,17 +126,30 @@ func (c *RunCommand) Run(ctx context.Context) error {
 		return fmt.Errorf("detecting cluster version: %w", err)
 	}
 
-	return c.runMigrationMode(ctx, currentVersion, c.parsedTargetVersion, c.registry)
+	effectivePhase, resolvedIDs := resolvePhaseAndMigrations(
```
chambridge: Are we running phase filtering twice? I see it being called in command_prepare.go, just wondering if this is redundant.

chambridge (commented):
Some feedback, but mostly looks good.

pkg/migrate/phase.go
asmigala: Added a warning when requesting explicit migration with mismatched phase

pkg/migrate/command_prepare.go
```diff
@@ -114,20 +137,45 @@ func (c *PrepareCommand) Run(ctx context.Context) error {
...
+	}
+
+	return c.runPrepareMode(ctx, currentVersion, c.parsedTargetVersion, c.registry, effectivePhase)
+}
+
+func (c *PrepareCommand) runPrepareMode(
+	ctx context.Context,
+	currentVersion *semver.Version,
+	targetVersion *semver.Version,
+	registry *action.ActionRegistry,
```
davidebianchi: nit: here the registry could be taken from `PrepareCommand` instead?

pkg/migrate/phase.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/opendatahub-io/odh-cli/pkg/util/client"
+	"github.com/opendatahub-io/odh-cli/pkg/util/iostreams"
+)
+
+// DetectPhase determines the migration phase based on version comparison.
+// If current < target: pre-upgrade (cluster hasn't been upgraded yet).
+// If current >= target: post-upgrade (cluster has been upgraded).
+func DetectPhase(current, target *semver.Version) action.ActionPhase {
+	if current == nil || target == nil {
+		return action.PhasePreUpgrade
```
davidebianchi: `DetectPhase` defaults to `PhasePreUpgrade` when either version is nil. It might be worth documenting *why* pre-upgrade is chosen as the default

pkg/migrate/command_run.go
```diff
@@ -126,6 +161,11 @@ func (c *RunCommand) runMigrationMode(
 			return fmt.Errorf("migration %q not found", migrationID)
 		}
 
+		if effectivePhase != "" && selectedAction.Phase() != effectivePhase {
```
davidebianchi: The `effectivePhase` variable is always non-empty at this point (because `resolvePhaseAndMigrations` always assigns a detected phase when the input phase is empty). This means the guard `effectivePhase != ""` is always true. Could the condition be simplified to just `selectedAction.Phase() != effectivePhase`.

pkg/migrate/phase.go
```diff
@@ -0,0 +1,78 @@
...
+
+	return action.PhasePostUpgrade
+}
+
+// resolvePhaseAndMigrations determines the effective phase and populates migration IDs
+// when they aren't explicitly provided. Returns the effective phase and the migration IDs.
+//
+// When --migration is explicit, phase filtering is bypassed: the migration runs
+// regardless of phase mismatch, with a warning emitted by the caller.
+func resolvePhaseAndMigrations(
```
davidebianchi: The `resolvePhaseAndMigrations` function always auto-detects a phase when `--phase` is omitted (`effectivePhase` will never be `""` after this block). This means the phase check inside `runMigrationMode` / `runPrepareMode` (`if effectivePhase != "" && selectedAction.Phase() != effectivePhase`) is always true, so phase filtering happens twice: once here via `filterByPhase` when resolving IDs, and again in the execution loop.

Consider having `resolvePhaseAndMigrations` be the single source of truth for phase filtering and removing the redundant check from the execution loops, or vice versa.

asmigala: The check in the run and prepare commands is there to emit a warning when a migration is selected explicitely but its phase doesn't match the effective phase.

pkg/migrate/phase.go
asmigala: That's actually a good point, maybe it would make more sense to error out here, wdyt?

davidebianchi: Agree, if are both required for migration, it makes sense to return an error to the user

davidebianchi (commented):
There are coderabbit comments that should be reviewed

pkg/migrate/command_prepare.go
```diff
@@ -174,7 +221,14 @@ func (c *PrepareCommand) Run(ctx context.Context) error {
 		c.IO.Errorf("Dry-run complete. Run without --dry-run to create backups.")
 	} else {
 		c.IO.Errorf("All preparations completed successfully!")
-		c.IO.Errorf("Backups saved to: %s", c.OutputDir)
+
+		entries, _ := os.ReadDir(c.OutputDir)
```
davidebianchi: The error from `os.ReadDir` is silently discarded here. If the directory read fails for a reason other than "not exists" (e.g., permission denied), we would misleadingly tell the user that no backups were created.

We could handle this more defensively:

```go
entries, err := os.ReadDir(c.OutputDir)
if err != nil && !os.IsNotExist(err) {
	c.IO.Errorf("Warning: could not read backup directory: %v", err)
} else if len(entries) > 0 {
	c.IO.Errorf("Backups saved to: %s", c.OutputDir)
} else {
	c.IO.Errorf("No backups were created ...")
}
```

pkg/migrate/phase.go
```diff
@@ -0,0 +1,78 @@
...
+	c client.Client,
+	io iostreams.Interface,
+) (action.ActionPhase, []string) {
+	effectivePhase := parsedPhase
+	if effectivePhase == "" {
+		effectivePhase = DetectPhase(currentVersion, targetVersion)
+		io.Errorf("Auto-detected phase: %s", string(effectivePhase))
+	}
+
+	if len(migrationIDs) == 0 {
```
davidebianchi, line 61: The `resolvePhaseAndMigrations` function accepts a `client.Client` parameter (`c`) but it is only passed through to `act.CanApply(action.Target{Client: c, ...})`. The variable name `c` shadows the common Go convention of using `c` for the receiver. Consider renaming to `cl` or `k8sClient` for clarity, especially since this function already has many parameters.

More broadly, this function has 7 parameters -- could we consider grouping some of these into a struct (e.g., a `phaseResolverInput`) to keep the call sites readable?

pkg/migrate/command_run.go
```diff
@@ -67,6 +81,10 @@ func (c *RunCommand) Complete() error {
 	// Always enable verbose for migrate run (both dry-run and actual execution)
 	c.Verbose = true
 
+	if c.Phase != "" {
```
davidebianchi, line 78: When `c.Phase` is empty, `c.parsedPhase` retains its zero value from construction. However, if a `RunCommand` instance were ever reused (e.g., in tests), a previously set `parsedPhase` would leak. Consider unconditionally resetting:

```go
c.parsedPhase = ""
if c.Phase != "" {
    c.parsedPhase = action.ActionPhase(c.Phase)
}
```

Same applies to `PrepareCommand.Complete()` at the equivalent location.

pkg/migrate/phase.go
asmigala: @davidebianchi added that as separate commit for now, if that looks ok, I'll squash

chambridge:
@asmigala looks like you just need a quick rebase to fix the conflict and then this PR looks good to merge.

chambridge (approved):
LGTM

pkg/migrate/action/executor.go
```diff
@@ -37,15 +37,26 @@ func (e *Executor) ExecuteSelective(
...
-	actions, err := e.registry.ListByPattern(pattern, group)
+	actions, err := e.registry.ListByFilter(pattern, group, phase)
 	if err != nil {
 		return nil, fmt.Errorf("selecting actions: %w", err)
 	}
 
 	return e.executeActions(ctx, target, actions), nil
 }
 
+func (e *Executor) ExecuteByPhase(
```
davidebianchi: `ExecuteByPhase` is added and tested but not called from any command code in this PR. The commands resolve phase and migration IDs via `resolvePhaseAndMigrations` and then iterate explicitly.

Is this intended as a public API for future use, or could it be deferred until a consumer exists? Having unused public methods can create maintenance burden if the calling convention evolves before it gets used.

pkg/migrate/command_prepare.go
```diff
@@ -49,6 +52,17 @@ func NewPrepareCommand(streams genericiooptions.IOStreams) *PrepareCommand {
...
+	actions := c.registry.ListAll()
+	ids := make([]string, 0, len(actions))
+
+	for _, a := range actions {
+		ids = append(ids, a.ID())
+	}
+
+	return ids
+}
+
```
davidebianchi, line 58: Does makes sense to have the ActionIDs fn in the registry? It is reused in multiple places

davidebianchi (approved):
Added a couple of nit comments, but LGTM

PR #81, opened 2026-05-21
feat: add workbenches.upgrade-2x-to-3x migration action (RHOAIENG-61433)

jeffdyoung:
## Description

Converts the workbench container-name-fix portion of
`workbench-2.x-to-3.x-upgrade.sh` to a Go migration action
(`workbenches.upgrade-2x-to-3x`) for the odh-cli migrate framework.

Dashboard-managed notebooks in RHOAI 2.x can end up with a primary
container name that doesn't match the Notebook CR name. This breaks
accelerator injection and size selection in 3.x. The action detects
and fixes these mismatches.

### Features
- Dry-run support across prepare and run phases
- Non-stopped notebook detection with `--force-non-stopped` override
- Pre-migration backup of notebook specs grouped by namespace
- Backup failure halts mutation (no half-broken state)
- DeepCopy before mutation to prevent in-memory corruption on update failure
- Partial failure tracking and reporting

### New files (4)
```
pkg/migrate/actions/workbenches/upgrade/
  upgrade.go       — Action struct, helpers (backup, classify, fix)
  prepare_task.go  — Backup notebook specs before migration
  run_task.go      — List → filter → backup → fix container names
  upgrade_test.go  — 68 test functions, 91.1% coverage
```

### Modified files (3)
- `pkg/migrate/command_list.go` — Register action
- `pkg/migrate/command_prepare.go` — Register action
- `pkg/migrate/command_run.go` — Register action

Ref: RHOAIENG-61433

## Testing

- [x] Unit tests pass — 68 tests in 1 file, 91.1% coverage
- [x] `go build`, `go vet`, `gofmt`, `go test -race` all pass
- [x] Tested against live RHOAI 2.25.6 cluster (7 notebooks, dry-run)

Test coverage includes:
- Container name mismatch detection and fix
- Multi-container notebooks with infrastructure sidecars
- Custom images, volume mounts, env vars preserved through rename
- Non-stopped notebook handling (skip, force, dry-run)
- Backup writes grouped by namespace + backup failure halts mutation
- Failing list/update client error paths
- Dry-run across all mutation paths
- ActionConfigurer flag registration

## Review checklist

## Summary by CodeRabbit

* **New Features**
  * Added a Workbench upgrade action to migrate notebooks 2.x→3.x: detects and fixes container-name mismatches, preserves container image/env/volume mounts, supports backups, dry-run, and optional inclusion of non-stopped notebooks with interactive confirmation.
  * Action is registered and available in migrate prepare, run, and list commands.

* **Tests**
  * Extensive tests covering validation, execution, backup/error cases, dry-run, prompting, multi-namespace, and update behaviors.

### DIFF
--- a/pkg/migrate/actions/workbenches/upgrade/prepare_task.go
+++ b/pkg/migrate/actions/workbenches/upgrade/prepare_task.go
@@ -1,0 +1,70 @@
+package upgrade
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/opendatahub-io/odh-cli/pkg/migrate/action"
+	"github.com/opendatahub-io/odh-cli/pkg/migrate/action/result"
+)
+
+type prepareTask struct {
+	action *WorkbenchUpgradeAction
+}
+
+func (t *prepareTask) Validate(
+	ctx context.Context,
+	target action.Target,
+) (*result.ActionResult, error) {
+	step := target.Recorder.Child(
+		"validate-notebooks",
+		"Check for Notebook resources",
+	)
+
+	notebooks, err := t.action.listNotebooks(ctx, target)
+	if err != nil {
+		step.Complete(result.StepFailed, "Failed to list Notebooks: %v", err)
+	} else if len(notebooks) == 0 {
+		step.Complete(result.StepCompleted, "No Notebook instances found")
+	} else {
+		step.Complete(result.StepCompleted, "Found %d Notebook(s) across the cluster", len(notebooks))
+	}
+
+	return buildResult(target.Recorder)
+}
+
+func (t *prepareTask) Execute(
+	ctx context.Context,
+	target action.Target,
+) (*result.ActionResult, error) {
+	step := target.Recorder.Child(
+		"prepare-backup",
+		"Prepare workbench backup",
+	)
+
+	notebooks, err := t.action.listNotebooks(ctx, target)
+	if err != nil {
+		step.Complete(result.StepFailed, "Failed to list Notebooks: %v", err)
+
+		return buildResult(target.Recorder)
+	}
+
+	if len(notebooks) == 0 {
+		step.Complete(result.StepCompleted, "No Notebook instances found, nothing to back up")
+
+		return buildResult(target.Recorder)
+	}
+
+	if err := t.action.backupNotebooks(ctx, target, notebooks, step); err != nil {
+		r, buildErr := buildResult(target.Recorder)
+		if buildErr != nil {
+			return nil, buildErr
+		}
+
+		return r, fmt.Errorf("prepare backup failed: %w", err)
+	}
+
+	step.Complete(result.StepCompleted, "Backup phase complete")
+
+	return buildResult(target.Recorder)
+}
--- a/pkg/migrate/actions/workbenches/upgrade/run_task.go
+++ b/pkg/migrate/actions/workbenches/upgrade/run_task.go
@@ -1,0 +1,237 @@
+package upgrade
+
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/opendatahub-io/odh-cli/pkg/migrate/action"
+	"github.com/opendatahub-io/odh-cli/pkg/migrate/action/result"
+)
+
+type runTask struct {
+	action *WorkbenchUpgradeAction
+}
+
+func (t *runTask) Validate(
+	ctx context.Context,
+	target action.Target,
+) (*result.ActionResult, error) {
+	step := target.Recorder.Child(
+		"validate-prerequisites",
+		"Validate migration prerequisites",
+	)
+
+	notebooks, err := t.action.listNotebooks(ctx, target)
+	if err != nil {
+		step.Complete(result.StepFailed, "Failed to list Notebooks: %v", err)
+
+		return buildResult(target.Recorder)
+	}
+
+	if len(notebooks) == 0 {
+		step.Complete(result.StepCompleted, "No Notebook instances found, nothing to migrate")
+
+		return buildResult(target.Recorder)
+	}
+
+	stopped := t.action.handleNonStoppedNotebooks(target, notebooks, step)
+
+	mismatchCount := 0
+	checkErrCount := 0
+
+	for _, nb := range stopped {
+		hasMismatch, checkErr := hasContainerNameMismatch(nb)
+		if checkErr != nil {
+			step.Record("check-error",
+				fmt.Sprintf("Failed to inspect Notebook %s/%s: %v", nb.GetNamespace(), nb.GetName(), checkErr),
+				result.StepFailed)
+			checkErrCount++
+
+			continue
+		}
+
+		if hasMismatch {
+			mismatchCount++
+		}
+	}
+
+	if checkErrCount > 0 {
+		step.Complete(result.StepFailed,
+			"Failed to inspect %d Notebook(s); resolve validation errors first", checkErrCount)
+	} else if mismatchCount > 0 {
+		step.Complete(result.StepCompleted,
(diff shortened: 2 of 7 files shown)


pkg/migrate/actions/workbenches/upgrade/prepare_task.go
```diff
@@ -0,0 +1,91 @@
...
+		step.Complete(result.StepCompleted, "Found %d Notebook(s) across the cluster", len(notebooks))
+	}
+
+	rootRecorder, ok := target.Recorder.(action.RootRecorder)
+	if !ok {
+		return nil, errors.New("recorder is not a RootRecorder")
+	}
+
+	return rootRecorder.Build(), nil
+}
```
davidebianchi, line 34: The `rootRecorder` type-assertion + `Build()` pattern is repeated 13 times across `prepare_task.go` and `run_task.go`. This is a lot of boilerplate that makes the actual task logic harder to read.

Consider extracting a helper, for example:
```go
func buildResult(recorder action.StepRecorder) (*result.ActionResult, error) {
    rootRecorder, ok := recorder.(action.RootRecorder)
    if \!ok {
        return nil, errors.New("recorder is not a RootRecorder")
    }
    return rootRecorder.Build(), nil
}
```
Then each return site becomes `return buildResult(target.Recorder)`. This also reduces the risk of forgetting the type assertion in future methods.

pkg/migrate/actions/workbenches/upgrade/run_task.go
```diff
@@ -0,0 +1,260 @@
...
+
+		rootRecorder, ok := target.Recorder.(action.RootRecorder)
+		if !ok {
+			return nil, errors.New("recorder is not a RootRecorder")
+		}
+
+		return rootRecorder.Build(), nil
+	}
+
+	if target.OutputDir == "" {
```
davidebianchi: The `run` phase creates its own backup directory (with a timestamp-based default) and backs up notebooks before mutation. The `prepare` phase does the same thing.

This means a typical workflow of `migrate prepare` followed by `migrate run` will produce two independent backup directories. Is that intentional? If so, a comment explaining the rationale (run-phase backup as a safety net independent of prepare) would be helpful. If not, we might want to accept an `--output-dir` flag on the run command too, or skip the backup in run if the user already ran prepare.

jeffdyoung:
@coderabbitai review

jeffdyoung:
https://github.com/coderabbitai review

pkg/migrate/actions/workbenches/upgrade/upgrade.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// isStopped returns true if the notebook has the kubeflow-resource-stopped annotation.
+func isStopped(nb *unstructured.Unstructured) bool {
+	annotations := nb.GetAnnotations()
+	if annotations == nil {
+		return false
+	}
+
+	_, stopped := annotations[annotationKubeflowResourceStopped]
```
davidebianchi, line 108: It is enough the presence, or it count also the value?

pkg/migrate/actions/workbenches/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,1958 @@
...
+func TestWorkbenchUpgradeAction_CanApply(t *testing.T) {
+	v3 := semver.MustParse("3.0.0")
+
+	tests := []struct {
+		name          string
+		current       string
+		targetVersion *semver.Version
+		expected      bool
+	}{
+		{"applies for 2.25 -> 3.0", "2.25.0", &v3, true},
```
davidebianchi, line 240: Should we add a check for 3.3 and 3.5?

pkg/migrate/actions/workbenches/upgrade/run_task.go
jeffdyoung: I think "skip the backup in run if the user already ran prepare" is the right thing to do here. I'll go ahead and change, but happy to revisit if you (or others) want to go a different direction.

pkg/migrate/actions/workbenches/upgrade/upgrade.go
jeffdyoung: ~~Good catch — presence alone wasn't enough. The Dashboard sets this annotation to "odh-notebook-controller-lock" during notebook creation to prevent a race with the controller. That's not a real stop. Updated isStopped to exclude that value, matching how hasStopAnnotation works in the Dashboard frontend (frontend/src/pages/projects/notebook/utils.ts). Added a test case for it too.~~

pkg/migrate/actions/workbenches/upgrade/upgrade.go
```diff
@@ -0,0 +1,383 @@
...
+func isStopped(nb *unstructured.Unstructured) bool {
+	annotations := nb.GetAnnotations()
+	if annotations == nil {
+		return false
+	}
+
+	val, ok := annotations[annotationKubeflowResourceStopped]
+
+	return ok && val != notebookControllerLock
+}
```
davidebianchi: I checked the original code, and it seems to only validate the presence as was before? https://github.com/red-hat-data-services/rhoai-upgrade-helpers/blob/a9aa2e6c8b03f54ed4a1699c25d66e58cbf90652/workbenches/workbench-2.x-to-3.x-upgrade.sh#L472C1-L485C2

Not sure here, should we ask for confirmation from workbenches team?

pkg/migrate/actions/workbenches/upgrade/upgrade.go
jeffdyoung: Ignore my above comment..chatting with the workbench team to determine the right path....

pkg/migrate/actions/workbenches/upgrade/upgrade.go
jeffdyoung: @davidebianchi Had a chat with @andyatmiami and @harshad16 
We reached a consensus that checking for the presence of the "only stopped" annotation is sufficient for the current scope. Since the intention is to prevent user disruption, and resources in a "starting" state have no active users, they do not need to evaluate the specific value of the annotation.

andyatmiami: `lgtm` w.r.t. @jeffdyoung 's summary
- _but as its nearing end of day - I'll just throw **way more words at this**_ 😈 

w.r.t this annotation's values:

- `$(date -u +%Y-%m-%dT%H:%M:%SZ)`
    - Notebook CR (Workbench) was stopped via Dashboard
    - this is an obvious "this is stopped - nothing will disrupt user" scenario
    
- `odh-notebook-controller-lock`
    - Notebook CR (Workbench) is actually in the process of "starting"
        - `odh-notebook-controller` "hijacks" this annotation to prevent the `notebook-controller` from spawinng a `StatefulSet` until some ODH-specific resource initialization completes
        - ℹ️ this only happens for `CREATE` - not `UPDATE`
    - while technically this notebook is actually "starting" - not "stopped" in this case - the ultimate point here is that an end user **absolutely has no expectation/ability of relying on this Workbench at this time** - since a `StatefulSet` has not even been generated.  As such - its safe to treat this as a "stopped" notebook for the context of migration - as restarting this workload cannot be deemed "disruptive" - as no process/human could have been relying on it yet
    
    
The 3rd case I'll call out here - although not specified in any of _our code_ - is that a customer could be doing their own wonky stuff - such that the value of this annotation is defined but not one of the 2 values outlined above.  As this value is ultimately arbitrary (except for this wonky `CREATE` webhook `odh-notebook-controller-lock` "magic string") - any other value defined is more or less safe to treat identically to the `$(date -u +%Y-%m-%dT%H:%M:%SZ)` variant.. i.e. any arbitrary value indicates the Workbench is, in fact, _stopped_

carlkyrillos:
LGTM

chambridge (approved):
lgtm

PR #83, opened 2026-05-25
feat(migrate): ActionConfigurer flag interface

grdryn:
This is the 1st PR in a stack, before #84 and #85, in an attempt to make review easier (I'm not sure if it does or not, let me know). It can either be reviewed and merged as-is, or by merging one or other of the other two.

## Description

https://issues.redhat.com/browse/RHOAIENG-61437

The purpose of this one is to allow the dashboard URL redirect action proposed in #85 to accept `--redirect-url` and `--redirect-route-host` flags akin to the flags that were available on the [old upgrade helper script](https://github.com/red-hat-data-services/rhoai-upgrade-helpers/tree/a9aa2e6c8b03f54ed4a1699c25d66e58cbf90652/dashboard).

- Add optional ActionConfigurer interface allowing actions to declare their own CLI flags via AddFlags(*pflag.FlagSet)
- Wire flag registration in RunCommand and PrepareCommand
- Add tests for override behaviour and command flag registration

## Testing

## Review checklist

## Summary by CodeRabbit

* **New Features**
  * Migration actions can now define and register their own CLI flags, enabling action-specific configuration options for prepare and run commands.

* **Tests**
  * Added tests to verify action-specific flags are properly registered and available.

### DIFF
--- a/pkg/migrate/mock_action_test.go
+++ b/pkg/migrate/mock_action_test.go
@@ -1,0 +1,27 @@
+package migrate_test
+
+import (
+	"github.com/spf13/pflag"
+
+	"github.com/opendatahub-io/odh-cli/pkg/migrate/action"
+)
+
+type MockActionWithFlags struct {
+	action.Action
+}
+
+func (m *MockActionWithFlags) ID() string                { return "mock.action.flags" }
+func (m *MockActionWithFlags) Name() string              { return "Mock Action with Flags" }
+func (m *MockActionWithFlags) Description() string       { return "Mock" }
+func (m *MockActionWithFlags) Group() action.ActionGroup { return action.GroupMigration }
+
+func (m *MockActionWithFlags) CanApply(target action.Target) bool { return true }
+func (m *MockActionWithFlags) Prepare() action.Task               { return nil }
+func (m *MockActionWithFlags) Run() action.Task                   { return nil }
+
+func (m *MockActionWithFlags) AddFlags(fs *pflag.FlagSet) {
+	fs.String("mock-flag-1", "", "")
+	fs.String("mock-flag-2", "", "")
+}
+
+var _ action.ActionConfigurer = (*MockActionWithFlags)(nil)
--- a/pkg/migrate/action/action.go
+++ b/pkg/migrate/action/action.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/blang/semver/v4"
+	"github.com/spf13/pflag"
 
 	"github.com/opendatahub-io/odh-cli/pkg/migrate/action/result"
 	"github.com/opendatahub-io/odh-cli/pkg/util/client"
@@ -40,6 +41,13 @@ type Action interface {
 
 	// Run returns the Task for the migration execution phase.
 	Run() Task
+}
+
+// ActionConfigurer is an optional interface that Actions can implement to register
+// their own CLI flags. Commands call AddFlags during flag setup for any registered
+// action that implements this interface.
+type ActionConfigurer interface {
+	AddFlags(fs *pflag.FlagSet)
 }
 
 // Target holds all context needed for executing migration actions.
--- a/pkg/migrate/action/flags.go
+++ b/pkg/migrate/action/flags.go
@@ -1,0 +1,48 @@
+package action
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/pflag"
+)
+
+// RegisterActionFlags registers flags from all ActionConfigurer implementations in the registry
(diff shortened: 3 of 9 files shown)


lburgazzoli:
if I got this right, this is effectively the same as registering commands on the actual migrate subcommand, like multiple action could theoretically install the same flag

grdryn:
> if I got this right, this is effectively the same as registering commands on the actual migrate subcommand, like multiple action could theoretically install the same flag

@lburgazzoli yes, I was not sure where the balance should be between a central way of doing it versus having each action doing it a potentially different way. Do you think it would be better restricted to the dashboard action in #85?

pkg/migrate/mock_action_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+type MockActionWithFlags struct {
+	action.Action
+}
+
+func (m *MockActionWithFlags) ID() string                { return "mock.action.flags" }
+func (m *MockActionWithFlags) Name() string              { return "Mock Action with Flags" }
+func (m *MockActionWithFlags) Description() string       { return "Mock" }
+func (m *MockActionWithFlags) Group() action.ActionGroup { return action.GroupMigration }
+
```
chambridge, line 17: nit: The mock embeds a nil `action.Action` but doesn't implement `Phase()` — if a test ever hits `ListAll()` with phase filtering this would panic. Easy fix:
`func (m *MockActionWithFlags) Phase() action.ActionPhase { return action.PhasePreUpgrade }`

chambridge:
> > if I got this right, this is effectively the same as registering commands on the actual migrate subcommand, like multiple action could theoretically install the same flag
> 
> @lburgazzoli yes, I was not sure where the balance should be between a central way of doing it versus having each action doing it a potentially different way. Do you think it would be better restricted to the dashboard action in #85?

Claude's thoughts' on the manner:

> Flag collision risk (@lburgazzoli concern, CodeRabbit's concern) — All actions register flags on the shared *pflag.FlagSet. If two actions register a flag with the same
> name (e.g., both add --namespace), pflag panics at startup. With 1 registered action today (kueue/rhbok) this is safe, but as more actions land, it becomes fragile.
> Three mitigations to consider:
> - Namespace flags: --<action-id>-<flag-name>
> - Recover from panics in the registration loop
> - Document the constraint and catch collisions in tests
> 
> My take: For 2-3 actions this is fine. Flag collision would be caught immediately during development (it's a startup panic, not a runtime edge case). We can harden later if action count grows. Not a blocker.

Personally, I like the practical approach of unblocking for now (given time constraints). If we want to namespace flags that might be a simpler approach moving forward. We could open a follow up Jira on that.

lburgazzoli:
> > if I got this right, this is effectively the same as registering commands on the actual migrate subcommand, like multiple action could theoretically install the same flag
> 
> @lburgazzoli yes, I was not sure where the balance should be between a central way of doing it versus having each action doing it a potentially different way. Do you think it would be better restricted to the dashboard action in #85?

no was mainly interested to confirm my understanding,

lburgazzoli:
> 
> Personally, I like the practical approach of unblocking for now (given time constraints). If we want to namespace flags that might be a simpler approach moving forward. We could open a follow up Jira on that.

agree

pkg/migrate/mock_action_test.go
grdryn: Are you sure? I can't see a `Phase()` on the [interface](https://github.com/opendatahub-io/odh-cli/blob/main/pkg/migrate/action/action.go#L27). Might you be thinking of some other unmerged work in progress?

grdryn:
@lburgazzoli @chambridge thanks for the feedback, it's this kind of discussion that I was hoping to have when splitting this to a separate PR :)

I believe what you're saying is that it's acceptable as-is for now? I've added a new commit on top to try to make it more explicit and to catch collisions. Is there anything else that you'd like to see changed around this? Thanks.

chambridge (approved):
LGTM

pkg/migrate/mock_action_test.go
lburgazzoli: Just FYI, I remember that at the beginning, there were something related to the phase, like to distinguish between pre and post upgrade steps

grdryn: > Are you sure? I can't see a Phase() on the [interface](https://github.com/opendatahub-io/odh-cli/blob/main/pkg/migrate/action/action.go#L27). Might you be thinking of some other unmerged work in progress?

This didn't age well 🙃 A separate change was merged earlier today before this one, which added the changes that @chambridge was likely referring to. Fixed in #88

PR #86, opened 2026-05-26
RHOAIENG-54672 - feat(lint): add --from-stdin for JSON/YAML batch input

ShettyGaurav:
## Description

[RHOAIENG-54672](https://redhat.atlassian.net/browse/RHOAIENG-54672)

Adds `--from-stdin` flag to the lint command for JSON/YAML batch input, enabling scripted/automated workflows.

**Changes:**
- New `pkg/util/stdin` utility package (~44 lines) for reusable stdin parsing
- Lint command integration via `--from-stdin` flag
- Strict unmarshaling rejects unknown fields (catches typos)
- Stdin values override flag defaults; omitted fields keep defaults

**Usage:**
```bash
# JSON
echo '{"checks": ["components.*"], "severity": "warning"}' | kubectl odh lint --from-stdin

# YAML
kubectl odh lint --from-stdin < lint-config.yaml
```

## Testing

## Review checklist

## Follow-up

PRs will be created to add `--from-stdin` support to:
- `status` command
- `migrate` command

## Summary by CodeRabbit

* **New Features**
  * Added --from-stdin flag to configure the lint command via JSON/YAML from stdin.
  * Stdin-provided options (checks, severity, target version, verbosity, output format) override defaults but not explicit CLI flags.
  * Warns if --from-stdin is used while stdin appears to be a terminal; rejects empty or overly large stdin inputs.

* **Tests**
  * Expanded tests covering JSON/YAML parsing, validation errors, precedence behavior, and piped/stdin detection.

### DIFF
--- a/pkg/lint/command.go
+++ b/pkg/lint/command.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"slices"
 	"strings"
 
@@ -43,6 +44,7 @@ import (
 	"github.com/opendatahub-io/odh-cli/pkg/util/client"
 	clierrors "github.com/opendatahub-io/odh-cli/pkg/util/errors"
 	"github.com/opendatahub-io/odh-cli/pkg/util/iostreams"
+	"github.com/opendatahub-io/odh-cli/pkg/util/stdin"
 	"github.com/opendatahub-io/odh-cli/pkg/util/version"
 )
 
@@ -82,6 +84,9 @@ type Command struct {
 	// registry is the check registry for this command instance.
 	// Explicitly populated to avoid global state and enable test isolation.
 	registry *check.CheckRegistry
+
+	// flags stores the FlagSet for checking explicitly set flags
+	flags *pflag.FlagSet
 }
 
 // NewCommand creates a new Command with defaults.
@@ -161,6 +166,7 @@ func NewCommand(
 
 // AddFlags registers command-specific flags with the provided FlagSet.
 func (c *Command) AddFlags(fs *pflag.FlagSet) {
+	c.flags = fs // Store for checking explicitly set flags in applyStdinInput
 	fs.StringVar(&c.TargetVersion, "target-version", "", flagDescTargetVersion)
 	fs.StringVarP((*string)(&c.OutputFormat), "output", "o", string(OutputFormatTable), flagDescOutput)
 	fs.StringVar((*string)(&c.SeverityLevel), "severity", string(SeverityLevelInfo), flagDescSeverity)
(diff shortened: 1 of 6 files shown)


ShettyGaurav:
@coderabbitai review

ShettyGaurav:
@coderabbitai review

pkg/lint/command.go
```diff
@@ -180,13 +183,65 @@ func (c *Command) AddFlags(fs *pflag.FlagSet) {
...
+	}
+
+	// Merge stdin values into command options (stdin overrides defaults)
+	c.applyStdinInput(&input)
+
+	return nil
+}
+
+// applyStdinInput merges stdin configuration into command options.
+func (c *Command) applyStdinInput(input *StdinInput) {
```
dayakar349: Should stdin values take precedence over explicitly-set CLI flags too? Right now if a user runs echo '{"severity":"critical"}' | odh lint --from-stdin --severity warning, the stdin value wins silently. Was that intentional, or would it make sense to check pflag.FlagSet.Changed() so explicit flags take priority?

pkg/lint/command_options.go
```diff
@@ -19,6 +19,29 @@ import (
...
+
+	// Severity sets the minimum severity level (replaces --severity flag)
+	Severity string `json:"severity,omitempty" yaml:"severity,omitempty"`
+
+	// TargetVersion sets the target version for upgrade checks (replaces --target-version flag)
+	TargetVersion string `json:"targetVersion,omitempty" yaml:"targetVersion,omitempty"`
+
+	// Verbose enables verbose output (replaces --verbose flag)
+	Verbose bool `json:"verbose,omitempty" yaml:"verbose,omitempty"`
+
```
dayakar349, line 37: What happens if someone passes an invalid values via stdin, does it get caught cleanly downstream? Since SeverityLevel.Validate() and OutputFormat.Validate() already exist, would it be worth calling them in applyStdinInput to give a clearer error message tied to the stdin input?

pkg/util/stdin/stdin.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	return nil
+}
+
+// IsPiped returns true if the file is not a terminal (i.e., has piped data).
+// Use this to warn users who specify --from-stdin but forget to pipe input.
+func IsPiped(f *os.File) bool {
+	stat, err := f.Stat()
+	if err != nil {
```
dayakar349, line 46: Is the silent false return in IsPiped when Stat() fails intentional? It's fine given it only gates a warning, but a short inline comment explaining why the error is swallowed would help future readers.

pkg/lint/command.go
ShettyGaurav: Fixed. CLI flags now take precedence over stdin values. Added flagChanged() helper that checks pflag.FlagSet.Changed() - stdin values only apply when the flag wasn't explicitly set. Added test case to verify.

pkg/lint/command_options.go
ShettyGaurav: Fixed. applyStdinInput() now validates severity and output using their .Validate() methods before applying. Errors include "stdin input:" prefix for clarity. Also wrapped with ExitValidation so the suggestion shows "Check the command arguments and flags" instead of "Unexpected error".

pkg/util/stdin/stdin.go
ShettyGaurav: Added comment explaining the intentional behavior:
```
// Stat failure is rare; return false to assume terminal and show warning.
// This is safe: the warning is non-blocking and helps catch user mistakes.
```

ShettyGaurav:
@coderabbitai review

dayakar349 (approved):

PR #90, opened 2026-05-28
RHOAIENG-61431: Convert AI Pipelines migration scripts to Go actions

asmigala:
Add three migrate actions converting the AI Pipelines shell scripts (check_before_upgrade.sh, update_dsp_role.sh, post_upgrade_check.sh) into the rhai-cli migrate framework:

- ai-pipelines.pre-upgrade-check: captures DSPA pod health state, migrates v1alpha1 resources to v1 with retry/backoff, detects custom roles missing datasciencepipelinesapplications/api subresource
- ai-pipelines.update-dsp-role: auto-detects and patches custom RBAC roles, inferring verbs from existing route.openshift.io rules
- ai-pipelines.post-upgrade-check: compares post-upgrade pod health against saved baseline with polling for recovery

Also introduces a shared registry (registry.go) so action registration is defined once instead of duplicated across list/prepare/run commands.

## Testing

## Review checklist

## Summary by CodeRabbit

* **New Features**
  * Pre- and post-upgrade checks for AI Pipelines: pod-health snapshots, recovery polling, and comparison reporting.
  * Automatic detection and optional remediation of RBAC gaps for AI Pipelines.
  * API-version migration for AI Pipelines resources with dry-run support.

* **Chores**
  * Migrate commands now initialize a default action set by default.

* **Tests**
  * Added comprehensive unit tests covering migration, pod-health, role classification, and upgrade checks.

### DIFF
--- a/pkg/migrate/actions/aipipelines/dspa.go
+++ b/pkg/migrate/actions/aipipelines/dspa.go
@@ -1,0 +1,169 @@
+package aipipelines
+
+import (
+	"context"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/opendatahub-io/odh-cli/pkg/migrate/action"
+	"github.com/opendatahub-io/odh-cli/pkg/migrate/action/result"
+	"github.com/opendatahub-io/odh-cli/pkg/resources"
+	"github.com/opendatahub-io/odh-cli/pkg/util/client"
+)
+
+type dspaInfo struct {
+	Name      string
+	Namespace string
+}
+
+func listDSPAs(ctx context.Context, c client.Client, rt resources.ResourceType) ([]dspaInfo, error) {
+	return listDSPAsInNamespace(ctx, c, rt, "")
+}
+
+func listDSPAsInNamespace(ctx context.Context, c client.Client, rt resources.ResourceType, namespace string) ([]dspaInfo, error) {
+	list, err := c.Dynamic().Resource(rt.GVR()).
+		Namespace(namespace).
+		List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("listing %s: %w", rt.Kind, err)
+	}
+
+	dspas := make([]dspaInfo, 0, len(list.Items))
+
+	for _, item := range list.Items {
+		dspas = append(dspas, dspaInfo{
+			Name:      item.GetName(),
+			Namespace: item.GetNamespace(),
+		})
+	}
+
+	return dspas, nil
+}
+
+type migrateOpts struct {
+	DryRun bool
+}
+
+func migrateDSPAToV1(
+	ctx context.Context,
+	c client.Client,
+	dspa dspaInfo,
+	opts migrateOpts,
+) error {
+	gvrAlpha := resources.DataSciencePipelinesApplicationV1Alpha1.GVR()
+	gvrV1 := resources.DataSciencePipelinesApplicationV1.GVR()
+
+	obj, err := c.Dynamic().Resource(gvrAlpha).
+		Namespace(dspa.Namespace).
+		Get(ctx, dspa.Name, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("getting v1alpha1 DSPA %s/%s: %w", dspa.Namespace, dspa.Name, err)
+	}
+
+	cleanObj := obj.DeepCopy()
+	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "creationTimestamp")
+	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "generation")
+	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "managedFields")
+	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "uid")
+	unstructured.RemoveNestedField(cleanObj.Object, "status")
+
+	cleanObj.SetAPIVersion(resources.DataSciencePipelinesApplicationV1.APIVersion())
+	cleanObj.SetKind(resources.DataSciencePipelinesApplicationV1.Kind)
+
+	updateOpts := metav1.UpdateOptions{}
+	if opts.DryRun {
+		updateOpts.DryRun = []string{metav1.DryRunAll}
+	}
+
+	_, err = c.Dynamic().Resource(gvrV1).
+		Namespace(dspa.Namespace).
+		Update(ctx, cleanObj, updateOpts)
+	if err != nil {
+		return fmt.Errorf("applying v1 DSPA %s/%s: %w", dspa.Namespace, dspa.Name, err)
+	}
+
+	return nil
+}
+
+func migrateAllDSPAsToV1(
+	ctx context.Context,
+	c client.Client,
+	recorder action.StepRecorder,
+	opts migrateOpts,
+) error {
+	step := recorder.Child("migrate-v1alpha1-to-v1", "Migrate v1alpha1 DSPAs to v1")
+
+	v1alpha1DSPAs, err := listDSPAs(ctx, c, resources.DataSciencePipelinesApplicationV1Alpha1)
+	if err != nil {
+		step.Complete(result.StepFailed, "Failed to list v1alpha1 DSPAs: %v", err)
+
+		return err
+	}
+
+	if len(v1alpha1DSPAs) == 0 {
+		step.Complete(result.StepCompleted, "No v1alpha1 DSPAs found")
+
+		return nil
+	}
+
+	step.Record("detected", "Found %d v1alpha1 DSPA(s) to migrate", result.StepCompleted, len(v1alpha1DSPAs))
+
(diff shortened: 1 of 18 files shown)


pkg/migrate/actions/aipipelines/dspa.go
```diff
@@ -0,0 +1,166 @@
...
+
+	obj, err := c.Dynamic().Resource(gvrAlpha).
+		Namespace(dspa.Namespace).
+		Get(ctx, dspa.Name, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("getting v1alpha1 DSPA %s/%s: %w", dspa.Namespace, dspa.Name, err)
+	}
+
+	cleanObj := obj.DeepCopy()
+	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "creationTimestamp")
```
davidebianchi: Same convention concern: `unstructured.RemoveNestedField()` is used six times here (lines 63-68). The `jq.Transform()` utility from `pkg/util/jq` can handle this more idiomatically, e.g.:
```go
jq.Transform(cleanObj, `del(.metadata.creationTimestamp, .metadata.generation, .metadata.managedFields, .metadata.resourceVersion, .metadata.uid, .status)`)
```
This collapses six calls into one and aligns with the project's unstructured field access convention.

pkg/migrate/actions/aipipelines/postcheck.go
```diff
@@ -0,0 +1,195 @@
...
+) []podGroupComparison {
+	ticker := time.NewTicker(postUpgradePollInterval)
+	defer ticker.Stop()
+
+	deadline := time.After(postUpgradeTimeout)
+
+	for {
+		select {
+		case <-ctx.Done():
+			comparisons, _ := comparePodHealth(ctx, target.Client, preState)
```
davidebianchi: When `ctx.Done()` fires, `comparePodHealth` is called with an already-cancelled context (line 112). The Kubernetes API call inside `comparePodHealth` will almost certainly fail due to the cancelled context, and the error is silently discarded (`comparisons, _ := comparePodHealth(ctx, ...)`). This means the returned `comparisons` slice will be nil, and `recordComparisons` will record nothing -- the user sees no output explaining what happened.

Consider either:
1. Using `context.Background()` for the final comparison after cancellation, or
2. Recording a specific message when ctx is cancelled (e.g., "Operation cancelled, unable to complete final comparison").

pkg/migrate/actions/aipipelines/updatedsprole.go
```diff
@@ -0,0 +1,216 @@
...
+			"op":   "add",
+			"path": "/rules/-",
+			"value": map[string]any{
+				"apiGroups":     []string{dspaAPIGroup},
+				"resources":     []string{"datasciencepipelinesapplications/api"},
+				"resourceNames": []string{dspa.Name},
+				"verbs":         verbs,
+			},
+		},
+	}
```
davidebianchi: The validation after patching re-classifies the role using `classifyRole`, which checks `hasRouteAPIGroup && !hasDSPASubresource`. This validates that *some* DSPA subresource rule exists, but does not verify the specific DSPA name or verbs that were just patched -- unlike the original bash script which validates all three (API group, resource, resourceName, and verbs).

If there were already a DSPA subresource rule for a *different* DSPA, `classifyRole` would return `NeedsFix: false` even though the intended rule was never created. Consider a more targeted validation that checks for the specific resourceName and verbs.

pkg/migrate/actions/aipipelines/dspa.go
```diff
@@ -0,0 +1,166 @@
...
+		if err != nil {
+			return false, fmt.Errorf("re-listing v1alpha1 DSPAs: %w", err)
+		}
+
+		if len(remaining) == 0 {
+			return true, nil
+		}
+
+		for _, dspa := range remaining {
+			if migrateErr := migrateDSPAToV1(ctx, c, dspa, migrateOpts{}); migrateErr != nil {
```
jeffdyoung: > **Bug: retry loop ignores the `opts` parameter — DryRun is always false inside the retry**
>
> `migrateAllDSPAsToV1` receives `opts migrateOpts` but the retry loop passes `migrateOpts{}` instead of `opts`. This means if the code path ever reaches the retry loop during dry-run (it currently can't because of the early return at line 111–119), migrations would execute for real.
>
> More importantly, this is a readability/correctness trap — a future change could remove the early return and the retry loop would silently ignore DryRun.
>
> ```diff
> -if migrateErr := migrateDSPAToV1(ctx, c, dspa, migrateOpts{}); migrateErr != nil {
> +if migrateErr := migrateDSPAToV1(ctx, c, dspa, opts); migrateErr != nil {
> ```

pkg/migrate/actions/aipipelines/role_classifier.go
```diff
@@ -0,0 +1,148 @@
...
+
+		apiGroups, _ := extractStringSlice(rule, "apiGroups")
+		ruleResources, _ := extractStringSlice(rule, "resources")
+		verbs, _ := extractStringSlice(rule, "verbs")
+
+		for _, g := range apiGroups {
+			if g == "route.openshift.io" {
+				hasRouteAPIGroup = true
+				classification.RouteVerbs = verbs
+			}
```
jeffdyoung, line 65: > **Minor: `RouteVerbs` captures verbs from only the last matching rule**
>
> If a role has multiple rules with `apiGroups: ["route.openshift.io"]`, `classification.RouteVerbs` will be overwritten each iteration, keeping only the verbs from the last matching rule. The original shell script uses jq to find ANY rule with `route.openshift.io` and infers the update from that — but the shell script doesn't infer verbs at all (they're a CLI argument).
>
> Since verb inference is new behavior not in the original script, consider whether the intent is to merge verbs from all `route.openshift.io` rules, or if a single match is sufficient. In practice most roles have one such rule, so this may not matter, but documenting the choice would help.

pkg/migrate/actions/aipipelines/updatedsprole.go
```diff
@@ -0,0 +1,216 @@
...
+		"patch-"+role.Namespace+"-"+role.RoleName,
+		fmt.Sprintf("Patch role %s/%s", role.Namespace, role.RoleName),
+	)
+
+	// Resolve DSPAs in this namespace
+	dspas, err := listDSPAs(ctx, target.Client, resources.DataSciencePipelinesApplicationV1)
+	if err != nil {
+		step.Complete(result.StepFailed, "Failed to list DSPAs in %s: %v", role.Namespace, err)
+
+		return
```
jeffdyoung, line 100: > **Performance: `listDSPAs` is called cluster-wide once per role, then filtered to namespace**
>
> Each call to `patchRole` (line 97) does a cluster-wide `listDSPAs` across all namespaces, then filters to the role's namespace at line 107–110. If there are N roles needing fixes, this makes N cluster-wide List calls.
>
> Consider calling `listDSPAs` once in `Execute` and passing the results down, or scoping the List to the specific namespace:
> ```go
> // Scope to namespace instead of cluster-wide
> dspas, err := c.Dynamic().Resource(resources.DataSciencePipelinesApplicationV1.GVR()).
>     Namespace(role.Namespace).
>     List(ctx, metav1.ListOptions{})
> ```
>
> Not blocking, but worth considering for clusters with many namespaces.

jeffdyoung (commented):
Agree with the code rabbit suggestions, and left a few minor nits.

pkg/migrate/actions/aipipelines/pod_health.go
```diff
@@ -0,0 +1,281 @@
...
+	if err != nil {
+		return state, fmt.Errorf("reading state from %s: %w", path, err)
+	}
+
+	if err := json.Unmarshal(data, &state); err != nil {
+		return state, fmt.Errorf("unmarshaling state from %s: %w", path, err)
+	}
+
+	return state, nil
+}
```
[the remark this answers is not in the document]
jeffdyoung: @asmigala Apologies, one more thing... otherwise LGTM:

This coderabbit semantic point is valid — cache directories are for disposable data and this state must survive between runs. However, switching to `os.UserConfigDir()` doesn't fix the underlying problem.

This runs in an OpenShift container with arbitrary UIDs. Verified on a live ROSA cluster (RHOAI 2.25.6) with UID 1000690000:

```
UID: 1000690000
HOME: /
XDG_CACHE_HOME: UNSET

mkdir -p /.cache/odh-cli/ai-pipelines → Permission denied
mkdir -p /tmp/rhoai-upgrade-backup/ai_pipelines → SUCCESS
```

`os.UserConfigDir()` has the same failure — it returns `$HOME/.config` → `/.config` → also not writable.

The Dockerfile already provisions a path for exactly this case (line 130):
```dockerfile
RUN mkdir -p /tmp/rhoai-upgrade-backup && chmod 1777 /tmp/rhoai-upgrade-backup
```

That directory is now unused by the Go code. Simplest fix — use the path the Dockerfile already provisions:
```go
func defaultStatePath() (string, error) {
	return filepath.Join("/tmp/rhoai-upgrade-backup/ai_pipelines", stateFileName), nil
}
```

Or if `os.UserCacheDir()` / `os.UserConfigDir()` is preferred, add `ENV HOME=/tmp` to the Dockerfile.

chambridge (approved):
lgtm

davidebianchi:
@asmigala can you resolve conflicts?

PR #92, opened 2026-06-01
feat(components): add --from-stdin support for components(enable,disable) command 

ShettyGaurav:
## Description

Jira: [RHOAIENG-54672](https://redhat.atlassian.net/browse/RHOAIENG-54672)

Follow-up to #86, which added `--from-stdin` to the `lint` command with reusable `pkg/util/stdin` utilities.

This PR extends the same pattern to `components enable` and `components disable` commands, enabling batch operations via piped input for scripted/automated workflows.

**Features:**
- Enable/disable multiple components in a single command
- Accepts JSON or YAML input via stdin
- CLI flags take precedence over stdin values
- Supports hybrid mode: positional arg + stdin flags

**Example usage:**
```bash
# Batch enable via JSON
echo '{"components": ["dashboard", "kserve", "ray"]}' | odh components enable --from-stdin

# Batch disable via YAML file (components.yaml)
# components:
#   - dashboard
#   - kserve
# dryRun: true
# skipConfirm: true
odh components disable --from-stdin < components.yaml

# Hybrid: component from arg, flags from stdin
echo '{"dryRun": true, "skipConfirm": true}' | odh components enable dashboard --from-stdin
```

## Testing

## Review checklist

## Future work

A follow-up PR will add `--from-stdin` support to other commands as needed (e.g., `deps install`, `migrate run`).

## Summary by CodeRabbit

* **New Features**
  * Added a --from-stdin option to enable/disable for bulk component lists and options via JSON/YAML stdin; introduced structured stdin input support.

* **Bug Fixes**
  * Commands now accept 0 or 1 positional component when stdin is used, require at least one component overall, and only set the positional name when present.

* **Tests**
  * Added tests for stdin parsing, validation, error cases, and CLI-flag override behavior.

### DIFF
--- a/pkg/components/commands.go
+++ b/pkg/components/commands.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"strings"
 
 	"github.com/spf13/pflag"
 
@@ -13,7 +15,9 @@ import (
 	"github.com/opendatahub-io/odh-cli/pkg/cmd"
 	"github.com/opendatahub-io/odh-cli/pkg/constants"
 	"github.com/opendatahub-io/odh-cli/pkg/util/client"
+	clierrors "github.com/opendatahub-io/odh-cli/pkg/util/errors"
 	"github.com/opendatahub-io/odh-cli/pkg/util/iostreams"
+	"github.com/opendatahub-io/odh-cli/pkg/util/stdin"
 )
 
 var (
@@ -27,9 +31,17 @@ type EnableCommand struct {
 	ConfigFlags *genericclioptions.ConfigFlags
 	Client      client.Client
 
+	// ComponentName is the single component name from positional arg (legacy).
 	ComponentName string
-	DryRun        bool
-	Yes           bool
+	// ComponentNames is the list of components to enable (from stdin or single name).
+	ComponentNames []string
+
+	DryRun    bool
+	Yes       bool
+	FromStdin bool
+
+	// flags stores the FlagSet for checking explicit flag usage.
+	flags *pflag.FlagSet
 }
 
 // NewEnableCommand creates a new EnableCommand with defaults.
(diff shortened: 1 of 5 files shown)


pkg/components/commands.go
```diff
@@ -66,26 +98,87 @@ func (c *EnableCommand) Complete() error {
...
+	if err := stdin.Parse(c.IO.In(), &input); err != nil {
+		return fmt.Errorf("parsing stdin: %w", err)
+	}
+
+	return c.applyStdinInput(&input)
+}
+
+// applyStdinInput merges stdin configuration into command options.
+// Explicit CLI flags take precedence over stdin values.
+func (c *EnableCommand) applyStdinInput(input *StdinInput) error {
```
dayakar349, line 117: strings.TrimSpace is used to check for empty names but the name itself isn't trimmed before storing, so whitespace-padded names would pass validation but fail at the API.

pkg/components/commands.go
```diff
@@ -48,14 +60,34 @@ func (c *EnableCommand) SetComponentName(name string) {
...
 // Complete resolves derived fields after flag parsing.
 func (c *EnableCommand) Complete() error {
+	// Parse stdin configuration if --from-stdin is specified
+	if c.FromStdin {
+		if err := c.parseStdinConfig(); err != nil {
+			//nolint:wrapcheck // NewExitCodeError is a same-module constructor
+			return clierrors.NewExitCodeError(clierrors.ExitValidation, err)
+		}
+	}
+
```
dayakar349, line 85: When --from-stdin provides components and a positional arg is also given, the positional arg is silently ignored. Consider warning or erroring to avoid unexpected behavior.

ShettyGaurav: Fixed. Now returns an error if both are provided

pkg/components/commands.go
ShettyGaurav: Fixed. Now trimming names before storing

pkg/components/commands.go
```diff
@@ -66,26 +98,94 @@ func (c *EnableCommand) Complete() error {
 	return nil
 }
 
+// parseStdinConfig reads and applies configuration from stdin.
+func (c *EnableCommand) parseStdinConfig() error {
+	if f, ok := c.IO.In().(*os.File); ok && !stdin.IsPiped(f) {
+		c.IO.Errorf(warnStdinIsTerminal)
+	}
```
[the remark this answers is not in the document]
ShettyGaurav: Now uses %s
Addressed

pkg/components/commands.go
```diff
@@ -66,26 +98,94 @@ func (c *EnableCommand) Complete() error {
...
+		if err := MutateComponentState(ctx, MutateConfig{
+			IO:            c.IO,
+			Client:        c.Client,
+			ComponentName: componentName,
+			TargetState:   constants.ManagementStateManaged,
+			ActionVerb:    "enable",
+			DryRun:        c.DryRun,
+			SkipConfirm:   c.Yes,
+		}); err != nil {
+			c.IO.Errorf("Error enabling '%s': %v\n", componentName, err)
```
dayakar349: IOStreams.Errorf uses fmt.Fprintln which already appends a newline, the explicit \n here produces a double newline after each error. Remove the trailing \n from both calls (here and line 342).

ShettyGaurav: Fixed. Removed the trailing `\n`

dayakar349 (approved):

PR #94, opened 2026-06-05
feat(migrate): add --from-stdin flag and dynamic v1/v2 API detection

ShettyGaurav:
## Description

Add `--from-stdin` flag to `migrate run` command for batch/scripted migrations and fix dynamic v1/v2 API version detection.

Jira: [RHOAIENG-54672](https://redhat.atlassian.net/browse/RHOAIENG-54672)

### --from-stdin support

Enables running migrations in automated pipelines by accepting configuration via stdin instead of CLI flags.

- Accepts both JSON and YAML formats
- Supports fields: `migrations`, `targetVersion`, `phase`, `dryRun`, `skipConfirm`
- CLI flags take precedence over stdin values for flexibility
- Warns when stdin is a terminal (likely user error)

Example usage:
```bash
echo '{"migrations": ["kueue.rhbok.migrate"], "targetVersion": "3.0.0", "dryRun": true}' | \
  kubectl odh migrate run --from-stdin
```

### v1/v2 API detection fix

Previously, version detection hardcoded v2 API which failed on ODH/RHOAI 2.x clusters that only have v1 API.

- Uses Kubernetes Discovery API to detect available API version dynamically
- Tries v2 first, falls back to v1 if not available
- Handles mid-upgrade scenario: when v2 CRD exists but only v1 instance is present, correctly falls back to v1
- Changed `Detect()` parameter from `client.Reader` to `client.Client` to access Discovery API

## Testing

## Review checklist

## Summary by CodeRabbit

* **New Features**
  * Add --from-stdin support for migration runs to supply migrations, target version, phase, dry-run and confirm behavior via stdin.
  * Detect cluster versions from v2 APIs with automatic fallback to v1 when needed.

* **Improvements**
  * Explicit CLI flags now take precedence over stdin values; clearer validation and error reporting for stdin parsing.

* **Tests**
  * Comprehensive tests added for stdin handling and v2/v1 detection and upgrade scenarios.

### DIFF
--- a/pkg/util/version/sources.go
+++ b/pkg/util/version/sources.go
@@ -6,16 +6,91 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 
+	"github.com/opendatahub-io/odh-cli/pkg/resources"
 	"github.com/opendatahub-io/odh-cli/pkg/util/client"
 	"github.com/opendatahub-io/odh-cli/pkg/util/jq"
+	"github.com/opendatahub-io/odh-cli/pkg/util/kube/discovery"
 )
 
-// DetectFromDataScienceCluster attempts to detect version from DataScienceCluster resource
+// getSingletonWithDiscovery dynamically detects the available API version and retrieves
+// a singleton resource. Tries v2 first, falls back to v1.
+// This avoids code duplication for resources that exist in both v1 and v2 API versions.
+func getSingletonWithDiscovery(
+	ctx context.Context,
+	c client.Client,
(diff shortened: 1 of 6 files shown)


ShettyGaurav:
@coderabbitai review

pkg/util/version/sources.go
```diff
@@ -65,7 +141,7 @@ func DetectFromDSCInitialization(ctx context.Context, c client.Reader) (string,
```
dayakar349, line 141: Minor: DetectFromOLM signature widened from client.Reader to client.Client unnecessarily. The function only calls c.OLM() (which is on Reader), so the wider interface breaks segregation without benefit. The other two DetectFrom* functions genuinely need Client for Discovery-based v2 detection, but OLM detection doesn't.

dayakar349 (approved):

pkg/util/version/sources.go
ShettyGaurav: Addressed

dayakar349 (approved):

PR #96, opened 2026-06-08
feat(events): add --component flag to filter events by ODH component

ShettyGaurav:
Jira: [RHOAIENG-54646](https://redhat.atlassian.net/browse/RHOAIENG-54646)
Follow-up to #74

## Description

Adds `--component` flag to `kubectl odh events` to filter events by ODH component.

### Usage

```bash
# Filter by component
kubectl odh events --component kserve
kubectl odh events --component modelregistry

# Combine with other flags
kubectl odh events --component dashboard --type Warning
kubectl odh events --component ray --since 1h
```

### How it works
- Filters events by looking up each event's InvolvedObject and checking for `app.kubernetes.io/part-of` label
- Handles components with non-standard labels via override mapping (e.g., `modelregistry` -> `model-registry-operator`)
- Caches label lookups to avoid repeated API calls
- Includes shell completion for valid component names

## Testing

## Review checklist

## Summary by CodeRabbit

* **New Features**
  * Added a --component flag to filter events by Open Data Hub component; help now includes a "Filter by component" example.
  * Shell completion for the flag suggests valid component names and disables filename completion.

* **Improvements**
  * Events are now filtered by component label for more accurate ownership matching.
  * Added canonical component label handling for known components.

* **Tests**
  * Added unit tests for component validation, label resolution, kind-to-resource mapping, and event filtering.

### DIFF
--- a/pkg/events/filter.go
+++ b/pkg/events/filter.go
@@ -3,11 +3,18 @@ package events
 import (
 	"cmp"
 	"context"
+	"fmt"
 	"slices"
 	"time"
 
 	"github.com/opendatahub-io/opendatahub-operator/pkg/clusterhealth"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/opendatahub-io/odh-cli/pkg/resources"
 	clierrors "github.com/opendatahub-io/odh-cli/pkg/util/errors"
 )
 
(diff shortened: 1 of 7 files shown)


ShettyGaurav:
@coderabbitai review

pkg/events/filter.go
```diff
@@ -78,3 +85,102 @@ func sortEventsByTime(events []clusterhealth.EventInfo) {
...
+
+	if err != nil {
+		return nil, fmt.Errorf("getting %s %s/%s: %w", gvr.Resource, namespace, name, err)
+	}
+
+	return obj, nil
+}
+
+// kindToGVR maps Kubernetes Kind to GroupVersionResource.
+func kindToGVR(kind string) schema.GroupVersionResource {
```
den-rgb, line 172: kindToGVR only maps core/apps kinds. events referencing CRD objects (InferenceService, RayCluster, Notebook) return empty GVR and get silently excluded from --component results. consider documenting this in the flag help or adding common CRD kinds.

pkg/resources/component_crs.go
```diff
@@ -43,3 +48,22 @@ func GetComponentCR(name string) *ResourceType {
 
 	return &rt
 }
+
+// componentLabelOverrides maps component names to their actual label values
+// when they differ from the component name.
+//
+//nolint:gochecknoglobals // Static label mapping configuration
+var componentLabelOverrides = map[string]string{
```
den-rgb, line 60: componentLabelOverrides covers aipipelines and modelregistry but newer components (feastoperator, trustyai, etc.) haven't been verified. if any use a different part-of label value than their CLI name, --component will return empty results. consider adding a code comment or integration test to catch drift.

pkg/events/filter.go
```diff
@@ -78,3 +85,102 @@ func sortEventsByTime(events []clusterhealth.EventInfo) {
...
+		err error
+	)
+
+	if namespace != "" {
+		obj, err = c.Client.Dynamic().Resource(gvr).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
+	} else {
+		obj, err = c.Client.Dynamic().Resource(gvr).Get(ctx, name, metav1.GetOptions{})
+	}
+
+	if err != nil {
```
den-rgb: np: wrapping the error in getObject then checking apierrors.IsNotFound on the wrapped result in the caller works but is subtle. returning the raw err here and letting the caller wrap after type-checking would be more explicit.

pkg/events/filter.go
ShettyGaurav: Documented in flag help and code comment.

pkg/resources/component_crs.go
ShettyGaurav: Added code comment + TestAllComponentsHaveLabelMapping test to catch drift.

pkg/events/filter.go
ShettyGaurav: Done. Now getObject returns the raw error

pkg/resources/component_crs_test.go
```diff
@@ -49,6 +49,42 @@ func TestComponentCRResourceTypes(t *testing.T) {
...
+				t.Errorf("GetComponentLabelValue(%q) = %q, want %q", tt.component, got, tt.want)
+			}
+		})
+	}
+}
+
+// TestAllComponentsHaveLabelMapping ensures every component in ComponentCRResourceTypes
+// returns a non-empty label value from GetComponentLabelValue. This catches drift where
+// a new component is added but its label override (if needed) is missing.
+func TestAllComponentsHaveLabelMapping(t *testing.T) {
```
den-rgb: GetComponentLabelValue returns the component name itself when no override exists, so the result is always non-empty. consider asserting against an expected-label map so adding a new component forces the dev to explicitly state its label value.

ShettyGaurav: Addressed.

den-rgb (approved):
/lgtm

PR #97, opened 2026-06-08
Update Go version to 1.26

CiaraM100104:
## Description
This PR updates the Go version to 1.26, as well address some formatting issues that were presented by Go 1.26's stricter rules about non-constants.

JIRA: https://redhat.atlassian.net/browse/RHOAIENG-65496 

## Testing

## Review checklist

## Summary by CodeRabbit

* **Chores**
  * Updated Go language version to 1.26 in build configuration and dependencies.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/opendatahub-io/odh-cli
 
-go 1.25.7
+go 1.26
 
 require (
 	github.com/blang/semver/v4 v4.0.0
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build stage - use native platform for builder to avoid emulation
-FROM --platform=$BUILDPLATFORM registry.access.redhat.com/ubi9/go-toolset:1.25 AS builder
+FROM --platform=$BUILDPLATFORM registry.access.redhat.com/ubi9/go-toolset:1.26 AS builder
 
 # Build arguments for cross-compilation
 ARG TARGETOS


pkg/migrate/action/recorder_impl.go
```diff
@@ -131,7 +131,7 @@ func (r *stepRecorderImpl) Record(name string, messageFormat string, status resu
 	}
 
 	child := r.Child(name, message)
-	child.Complete(status, message)
+	child.Complete(status, "%s", message)
```
davidebianchi: Should be Completef in case to follow a well known pattern? Or, could the issue be solved in some other ways?

pkg/migrate/actions/aipipelines/postcheck.go
```diff
@@ -182,7 +182,7 @@ func (t *postUpgradeCheckRunTask) recordComparisons(
 			recordFailedPodGroup(groupStep, comp.Current)
 		}
 
-		groupStep.Complete(status, msg)
+		groupStep.Complete(status, "%s", msg)
```
davidebianchi: Same https://github.com/opendatahub-io/odh-cli/pull/97/changes#r3389436428

pkg/migrate/action/recorder_impl.go
jeffdyoung: Lets do the Completef change here first: https://github.com/opendatahub-io/odh-cli/pull/101

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/opendatahub-io/odh-cli
 
-go 1.25.7
+go 1.26
```
jeffdyoung, line 3: @grdryn We need some help here....Any reason we shouldn't roll with `1.26`? do we need to pin a specific version of `1.26.z`? Making sure I'm not missing something.

davidebianchi (approved):

go.mod
grdryn: @jeffdyoung x.y should be fine now, in my understanding.

We did hit an issue in the past with certain developer machines, when just specifying x.y versions, but it seems like that issue has been solved.

The issue was that if the developer's machine didn't have a new enough Go version, then go would fail to download the right toolchain, as follows:

```
➜ GOTOOLCHAIN=local go version
go version go1.21.3 darwin/amd64

➜ GOTOOLCHAIN=auto go version
go: downloading go1.22 (darwin/amd64)
go: download go1.22 for darwin/amd64: toolchain not available
```

so we'd have to specify 1.22.0, and it would then work:

```
➜ GOTOOLCHAIN=auto go version
go version go1.22.0 darwin/amd64
```

However, I've just checked on both Linux and MacOS on this PR with my default Go being 1.25, and it appears to work just fine:

```
➜ GOTOOLCHAIN=local go version
go version go1.25.11 darwin/arm64

➜ GOTOOLCHAIN=auto go version
go: downloading go1.26.0 (darwin/arm64)
go version go1.26.0 darwin/arm64
```

I'm not aware of any other reason to need to specify a full x.y.z version here.

grdryn (approved):

go.mod
grdryn: Related: https://github.com/golang/go/issues/65568

grdryn:
@CiaraM100104 @jeffdyoung could you please update the base image references in the downstream Dockerfile.konflux? Since the go.mod has been updated to 1.26 here, and that still uses 1.25 builder images, it's likely to fail (not sure why it hasn't already)

https://github.com/red-hat-data-services/odh-cli/blob/main/Dockerfile.konflux

CiaraM100104:
@grdryn Just created the pull request for the downstream update: https://github.com/red-hat-data-services/odh-cli/pull/287, let me know if there's anything else I need to do and thanks for letting me know!

PR #105, opened 2026-06-19
fix(migrate): check CRD storedVersions instead of listing via v1alpha1 GVR

asmigala:
## Description

The v1alpha1 DSPA migration always timed out because listing via the v1alpha1 GVR endpoint returns objects regardless of their stored version — the API server converts v1 objects to v1alpha1 on the fly. The retry loop could never observe "zero v1alpha1 DSPAs remaining."

Check the CRD's status.storedVersions instead. If v1alpha1 is not in storedVersions, all objects are already stored as v1 and the migration exits immediately. When migration is needed, verify by tracking Update success rather than re-listing via v1alpha1.

Ref: RHOAIENG-69646

## Testing

## Review checklist

## Release Notes

* **Refactor**
  * Updated DSPA migration detection to rely on CRD `storedVersions` (instead of enumerating v1alpha1 resources), including more resilient handling when CRD data is unavailable.
  * Improved migration retries to follow the initially selected DSPAs and avoid re-scanning on each attempt.
  * After successful migration, the CRD `storedVersions` is explicitly cleaned up to remove the v1alpha1 entry.

* **Tests**
  * Added unit tests covering CRD `storedVersions` detection and removing the v1alpha1 stored version.

### DIFF
--- a/pkg/migrate/actions/aipipelines/dspa.go
+++ b/pkg/migrate/actions/aipipelines/dspa.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -12,6 +13,7 @@ import (
 	"github.com/opendatahub-io/odh-cli/pkg/migrate/action/result"
 	"github.com/opendatahub-io/odh-cli/pkg/resources"
 	"github.com/opendatahub-io/odh-cli/pkg/util/client"
+	"github.com/opendatahub-io/odh-cli/pkg/util/jq"
 )
 
 type dspaInfo struct {
@@ -53,25 +55,20 @@ func migrateDSPAToV1(
 	dspa dspaInfo,
 	opts migrateOpts,
 ) error {
-	gvrAlpha := resources.DataSciencePipelinesApplicationV1Alpha1.GVR()
 	gvrV1 := resources.DataSciencePipelinesApplicationV1.GVR()
 
-	obj, err := c.Dynamic().Resource(gvrAlpha).
+	obj, err := c.Dynamic().Resource(gvrV1).
 		Namespace(dspa.Namespace).
 		Get(ctx, dspa.Name, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("getting v1alpha1 DSPA %s/%s: %w", dspa.Namespace, dspa.Name, err)
+		return fmt.Errorf("getting DSPA %s/%s: %w", dspa.Namespace, dspa.Name, err)
 	}
 
-	cleanObj := obj.DeepCopy()
-	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "creationTimestamp")
-	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "generation")
-	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "managedFields")
-	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "uid")
-	unstructured.RemoveNestedField(cleanObj.Object, "status")
-
-	cleanObj.SetAPIVersion(resources.DataSciencePipelinesApplicationV1.APIVersion())
-	cleanObj.SetKind(resources.DataSciencePipelinesApplicationV1.Kind)
+	unstructured.RemoveNestedField(obj.Object, "metadata", "creationTimestamp")
+	unstructured.RemoveNestedField(obj.Object, "metadata", "generation")
+	unstructured.RemoveNestedField(obj.Object, "metadata", "managedFields")
+	unstructured.RemoveNestedField(obj.Object, "metadata", "uid")
+	unstructured.RemoveNestedField(obj.Object, "status")
 
 	updateOpts := metav1.UpdateOptions{}
 	if opts.DryRun {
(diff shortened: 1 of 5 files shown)


pkg/migrate/actions/aipipelines/dspa.go
```diff
@@ -129,32 +181,20 @@ func migrateAllDSPAsToV1(
...
-		}
-
-		if len(remaining) == 0 {
-			return true, nil
-		}
+		allSucceeded := true
 
-		for _, dspa := range remaining {
+		for _, dspa := range dspas {
 			if migrateErr := migrateDSPAToV1(ctx, c, dspa, opts); migrateErr != nil {
```
davidebianchi, line 247: For the migrateDSPAToV1, could it be simplified by reading the default resource version (which should be v1) and patch it with the same content? In this way the migration will be done server side. 
It would be similar of today, but without the get in gvrAlpha and update in v1

pkg/migrate/actions/aipipelines/dspa.go
```diff
@@ -129,32 +181,20 @@ func migrateAllDSPAsToV1(
...
 			}
 		}
 
-		remaining, err = listDSPAs(ctx, c, resources.DataSciencePipelinesApplicationV1Alpha1)
-		if err != nil {
-			return false, fmt.Errorf("verifying migration: %w", err)
-		}
-
-		return len(remaining) == 0, nil
+		return allSucceeded, nil
```
davidebianchi, line 258: Once migration completed for all resources, should we remove the stored version from CRD?

pkg/migrate/actions/aipipelines/dspa.go
```diff
@@ -96,27 +169,44 @@ func migrateAllDSPAsToV1(
...
+	if err != nil {
+		step.Completef(result.StepFailed, "Failed to list DSPAs: %v", err)
+
+		return err
+	}
+
+	if len(dspas) == 0 {
+		step.Completef(result.StepCompleted, "No DSPAs found")
+
+		return nil
```
davidebianchi, line 213: When `hasV1Alpha1StoredVersion` returns `true` but no DSPAs exist on the cluster, the function exits early here without cleaning up the stale `v1alpha1` entry in `storedVersions`. Every subsequent run will re-enter the migration path (list DSPAs, find zero, exit) instead of taking the fast "All DSPAs already stored as v1" path.

This is not harmful, but it is wasteful and produces misleading output. We could call `removeV1Alpha1StoredVersion` before returning, since there are genuinely no v1alpha1 objects left to worry about:

```go
if len(dspas) == 0 {
    if \!opts.DryRun {
        if err := removeV1Alpha1StoredVersion(ctx, c); err \!= nil {
            step.Completef(result.StepFailed, "Failed to clean up storedVersions: %%v", err)
            return fmt.Errorf("removing v1alpha1 from storedVersions: %%w", err)
        }
    }
    step.Completef(result.StepCompleted, "No DSPAs found")
    return nil
}
```

pkg/migrate/actions/aipipelines/dspa.go
```diff
@@ -53,36 +55,107 @@ func migrateDSPAToV1(
...
-		return fmt.Errorf("getting v1alpha1 DSPA %s/%s: %w", dspa.Namespace, dspa.Name, err)
+		return fmt.Errorf("getting DSPA %s/%s: %w", dspa.Namespace, dspa.Name, err)
 	}
 
-	cleanObj := obj.DeepCopy()
-	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "creationTimestamp")
-	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "generation")
-	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "managedFields")
-	unstructured.RemoveNestedField(cleanObj.Object, "metadata", "uid")
-	unstructured.RemoveNestedField(cleanObj.Object, "status")
```
davidebianchi, line 71: Are those not needed?

pkg/migrate/actions/aipipelines/dspa.go
```diff
@@ -163,6 +241,14 @@ func migrateAllDSPAsToV1(
 		return fmt.Errorf("migrating v1alpha1 DSPAs: %w", retryErr)
 	}
 
+	if !opts.DryRun {
```
davidebianchi: If it's in dry run mode, isn't returned at line 211?

pkg/migrate/actions/aipipelines/dspa_test.go
```diff
@@ -94,3 +115,86 @@ func TestListDSPAs(t *testing.T) {
...
+		crd := makeDSPACRD("v1alpha1", "v1")
+		c := newFakeClient(crd)
+
+		err := aipipelines.RemoveV1Alpha1StoredVersion(ctx, c)
+		g.Expect(err).ToNot(HaveOccurred())
+
+		has, err := aipipelines.HasV1Alpha1StoredVersion(ctx, c)
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(has).To(BeFalse())
+	})
```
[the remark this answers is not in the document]
ugiordan: This is a valid suggestion.

pkg/migrate/actions/aipipelines/dspa.go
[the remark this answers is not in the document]
ugiordan: This is valid.

pkg/migrate/actions/aipipelines/dspa.go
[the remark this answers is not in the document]
ugiordan: @coderabbitai this is a better solution https://github.com/opendatahub-io/odh-cli/pull/105#discussion_r3443076511

pkg/migrate/actions/aipipelines/dspa.go
asmigala: @davidebianchi I think no, since we are just reapplying the same resource we got, but maybe it's safer to remove them?

ugiordan (approved):

davidebianchi (approved):

PR #107, opened 2026-06-25
[RHOAIENG-66868] feat(wait): add centralized --wait-for polling for operation chaining

ShettyGaurav:
## Description

Adds `--wait-for=healthy` polling to `kubectl odh status`, allowing users and automation to block until the platform reaches a desired state.

**Jira:** [RHOAIENG-66868](https://redhat.atlassian.net/browse/RHOAIENG-66868)

### What this PR does

- Introduces a centralized `WaitOptions`  in `pkg/cmd/` that any command can embed to get `--wait-for` and `--poll-interval` support
- Wires `--wait-for=healthy` into the status command with configurable `--timeout` and `--poll-interval`

### Usage

```bash
# Wait until healthy (5 min timeout)
kubectl odh status --wait-for=healthy --timeout=300s

# Custom poll interval
kubectl odh status --wait-for=healthy --timeout=300s --poll-interval=5s

# Wait indefinitely
kubectl odh status --wait-for=healthy --timeout=0
```

### Follow-up PR

`--watch`: stream newline-delimited JSON per state change with informer-style reconnection on network interruption (tracked in the same Jira).

## Testing

- [x] Validated against live ODH Cluster
## Review checklist

- [x] Centralized `WaitOptions` in `pkg/cmd/` for reuse by future commands
- [x] Structured errors for all validation and timeout cases

## Summary by CodeRabbit

* **New Features**
  * Added wait-mode support to the status command with `--wait-for=healthy`, configurable `--poll-interval`, and timeout behavior.
* **Documentation**
  * Expanded `status` command help/examples for healthy waiting, including default `--timeout=30s`, `--timeout=0` (no timeout), and timeout exit code `5`.
* **Bug Fixes**
  * Strengthened wait validation (poll interval and timeout bounds) and improved waiting progress/retry messaging, including structured timeout errors.
* **Tests**
  * Added/expanded test coverage for wait validation, wait execution, and end-to-end healthy-wait behavior.

### DIFF
--- a/pkg/cmd/wait.go
+++ b/pkg/cmd/wait.go
@@ -1,0 +1,127 @@
+package cmd
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+	"strings"
+	"time"
+
+	"github.com/spf13/pflag"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/opendatahub-io/odh-cli/pkg/api"
+)
+
+const (
+	DefaultPollInterval = 10 * time.Second
+	MaxPollInterval     = 5 * time.Minute
+	MaxWaitTimeout      = 30 * time.Minute
+
+	flagDescWaitFor      = `Wait for a condition before exiting`
+	flagDescPollInterval = "Poll interval for --wait-for (lower values increase API server load)"
+)
+
+// WaitCondition is a function that polls and returns (done, error).
+type WaitCondition func(ctx context.Context) (done bool, err error)
+
+// enumValue implements pflag.Value for a string constrained to a fixed set.
+type enumValue struct {
+	allowed []string
+	value   *string
+}
+
+func newEnumValue(allowed []string, target *string) *enumValue {
+	return &enumValue{allowed: allowed, value: target}
+}
+
+func (e *enumValue) String() string { return *e.value }
+
+func (e *enumValue) Set(s string) error {
+	if !slices.Contains(e.allowed, s) {
+		return fmt.Errorf("must be one of: %s", strings.Join(e.allowed, ", "))
+	}
+
+	*e.value = s
+
+	return nil
+}
+
+func (e *enumValue) Type() string {
+	return strings.Join(e.allowed, "|")
+}
+
+// WaitOptions provides reusable --wait-for and --poll-interval flag
+// support. Commands embed this struct and call its methods from their
+// own AddFlags, Validate, and Run implementations.
+type WaitOptions struct {
+	WaitFor      string
+	PollInterval time.Duration
+}
+
+// AddWaitFlags registers --wait-for and --poll-interval flags.
+// validConditions lists the values accepted by --wait-for; invalid
+// values are rejected at parse time by the enum flag type.
+func (w *WaitOptions) AddWaitFlags(fs *pflag.FlagSet, validConditions []string) {
+	fs.Var(newEnumValue(validConditions, &w.WaitFor), "wait-for", flagDescWaitFor)
+	_ = fs.SetAnnotation("wait-for", api.AnnotationValidValues, validConditions)
+	fs.DurationVar(&w.PollInterval, "poll-interval", w.PollInterval, flagDescPollInterval)
+}
+
+// HasWaitMode returns true if --wait-for is active.
+func (w *WaitOptions) HasWaitMode() bool {
+	return w.WaitFor != ""
+}
+
+// ValidateWait checks that --poll-interval and timeout are within bounds.
+// Condition validation is handled at parse time by the enum flag type.
+// timeout is the caller's --timeout value (0 means no timeout in wait mode).
+func (w *WaitOptions) ValidateWait(timeout time.Duration) error {
+	if w.WaitFor == "" {
+		return nil
+	}
+
+	if w.PollInterval < time.Second || w.PollInterval > MaxPollInterval {
+		return ErrInvalidPollInterval()
+	}
+
+	if timeout < 0 || timeout > MaxWaitTimeout {
+		return ErrInvalidWaitTimeout()
+	}
+
+	return nil
+}
+
+// RunWait polls cond at the configured interval until it returns true,
+// the timeout expires, or the context is cancelled. Pass timeout=0 for
+// no timeout (poll until condition or cancellation).
+func (w *WaitOptions) RunWait(ctx context.Context, timeout time.Duration, cond WaitCondition) error {
+	if w.PollInterval < time.Second || w.PollInterval > MaxPollInterval {
+		return ErrInvalidPollInterval()
+	}
+
+	if timeout < 0 {
+		return ErrInvalidWaitTimeout()
+	}
+
+	if timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, timeout)
+
+		defer cancel()
+	}
+
+	pollErr := wait.PollUntilContextCancel(ctx, w.PollInterval, true, wait.ConditionWithContextFunc(cond))
+
+	if pollErr != nil {
+		if errors.Is(pollErr, context.DeadlineExceeded) {
+			return ErrWaitTimeout(w.WaitFor)
+		}
+
+		return fmt.Errorf("waiting for condition %q: %w", w.WaitFor, pollErr)
+	}
+
+	return nil
+}
--- a/pkg/cmd/wait_errors.go
+++ b/pkg/cmd/wait_errors.go
@@ -1,0 +1,54 @@
+package cmd
+
+import (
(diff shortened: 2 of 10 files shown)


ShettyGaurav:
@coderabbitai review

pkg/cmd/wait.go
```diff
@@ -0,0 +1,93 @@
...
+	if pollErr != nil {
+		if errors.Is(pollErr, context.DeadlineExceeded) {
+			return ErrWaitTimeout(w.WaitFor)
+		}
+
+		return fmt.Errorf("waiting for condition %q: %w", w.WaitFor, pollErr)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
ShettyGaurav: Fixed. Added defensive guard at the top of RunWait

pkg/cmd/wait.go
```diff
@@ -0,0 +1,97 @@
...
+func (w *WaitOptions) ValidateWait(timeout time.Duration, validConditions []string) error {
+	if w.WaitFor == "" {
+		return nil
+	}
+
+	if !slices.Contains(validConditions, w.WaitFor) {
+		return ErrInvalidWaitCondition(w.WaitFor, validConditions)
+	}
+
+	if w.PollInterval <= 0 {
```
macgregor: nit: thought exercise for both poll interval and timeout, do we need to worry about an upper bound as well? Probably not, but worth giving it a think to see if there are consequences to someone just wails on the keyboard

pkg/cmd/wait_errors.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// ErrInvalidWaitCondition creates a structured error for unknown --wait-for values.
+func ErrInvalidWaitCondition(condition string, validConditions []string) *clierrors.StructuredError {
+	return &clierrors.StructuredError{
+		Code:       errCodeInvalidWaitCondition,
+		Message:    fmt.Sprintf(msgInvalidWaitCondition, condition),
+		Category:   clierrors.CategoryValidation,
+		Retriable:  false,
+		Suggestion: "Valid conditions: " + strings.Join(quotedConditions(validConditions), ", "),
```
macgregor: Great stuff, thorough error handling, with actionable wording is one of the primary foundations of good software.

pkg/status/config.go
```diff
@@ -0,0 +1,96 @@
...
+		if err != nil {
+			return fmt.Errorf("health checks: %w", err)
+		}
+
+		return nil
+	})
+
+	if c.IncludeDeps {
+		g.Go(func() error {
+			depStatuses = c.checkDependencies(gctx)
```
macgregor, line 92: does this not have potential to error? if so you might be swallowing them here.

pkg/status/status.go
```diff
@@ -122,6 +132,7 @@ func (c *Command) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.OperatorName, "operator-name", "", flagDescOperName)
 	fs.BoolVar(&c.IncludeInfra, "include-infra", false, flagDescInfra)
 	fs.BoolVar(&c.IncludeDeps, "include-deps", c.IncludeDeps, flagDescIncludeDeps)
+	c.AddWaitFlags(fs, []string{WaitConditionHealthy})
```
macgregor, line 136: nit: this is the sort of case where i wish Golang properly had enums as first class citizens. This would be a natural use case for them to easily represent "it has to be one of these few things". Basically, a more strongly typed representation would let you write less scaffolding and error handling code than simple strings (let the compiler and pflag library do the heavy lifting). isolated to this wait-for case alone, its probably more effort than warranted. Our other cli commands likely do the same sort of thing youre doing here, so its probably best just to stick with this pattern. 

Strong typing is our friend, even more so in this gen-AI world - strong typing gives you a feedback loop, the compiler can tell the AI it did something wrong rather than relying on a generated test to catch a generated problem. I'll leave it to you to decide if you want to start establishing a "better way" so to speak.

```
pflag doesn't have a built-in enum type, but its pflag.Var + custom Value interface gets you exactly that -- validation at parse time and allowed values in help text:

type enumValue struct {
    allowed []string
    value   *string
}

func newEnumValue(allowed []string, target *string) *enumValue {
    return &enumValue{allowed: allowed, value: target}
}

func (e *enumValue) String() string { return *e.value }

func (e *enumValue) Set(s string) error {
    if !slices.Contains(e.allowed, s) {
        return fmt.Errorf("must be one of: %s", str
    }
    *e.value = s
    return nil
}

func (e *enumValue) Type() string {
    return strings.Join(e.allowed, "|")
}

Then registration becomes:

func (w *WaitOptions) AddWaitFlags(fs *pflag.FlagSe{
    fs.Var(newEnumValue(validConditions, &w.WaitFor), "wait-for", flagDescWaitFor)
    fs.DurationVar(&w.PollInterval, "poll-interval", w.PollInterval, flagDescPollInterval)                                           }

Help output automatically shows --wait-for ready|complete|deleted (whatever Type() returns), and Set() rejects invalid values during parsing. That eliminates ValidateWait's condition cll -- both are now handled by the flag itself.
```

pkg/status/status.go
```diff
@@ -90,6 +94,9 @@ type Command struct {
 	// Populated during Complete
 	restConfig *rest.Config
 	client     client.Client
+
+	// healthCheckFn overrides buildHealthConfig+runHealthCheck for testing.
+	healthCheckFn func(ctx context.Context) (*clusterhealth.Report, []deps.DependencyStatus, error)
```
macgregor: `healthCheckFn` test seam bypasses testable code. clusterhealth.Config.Client is a controller-runtime client.Client interface, fakeable with fake.NewClientBuilder().Build(). Tests could inject a *clusterhealth.Config with a fake client instead of replacing the entire health check function. This would let tests exercise runHealthCheck and clusterhealth.Run rather than skipping over them. 

Concrete change: replace healthCheckFn func(...) field with healthConfig *clusterhealth.Config -- nil means build from cluster at runtime, non-nil means use as-is.

pkg/status/config.go
```diff
@@ -0,0 +1,96 @@
...
+	var report *clusterhealth.Report
+	var depStatuses []deps.DependencyStatus
+
+	g, gctx := errgroup.WithContext(ctx)
+
+	g.Go(func() error {
+		var err error
+		report, err = clusterhealth.Run(gctx, *hc)
+		if err != nil {
+			return fmt.Errorf("health checks: %w", err)
```
macgregor: nit: Unnecessary error wrapping adds noise. runHealthCheck wraps with "health checks:", then Run wraps again with "running status checks:". Since only one goroutine can return an error (dependency check always returns nil), the inner wrap doesn't help distinguish sources. The test file also has to manually reconstruct this wrapping chain for errForbidden. Consider dropping the wrappers, i.e. just `return clusterhealth.Run(gctx, *hc))` and in the `g.Wait()` block just `return return nil, nil, err)`. 

If you do end up finding errors you can bubble up from checkDependencies, you should still consider if the error wrapping adds value or if you can simply return the error as is.

macgregor (commented):
good work. A few thoughts for you to consider, none of them large enough to block anything. Do what you will with them and then i think its ready to merge.

pkg/cmd/wait.go
ShettyGaurav: Addressed : Added upper bounds: MaxPollInterval = 5m, MaxWaitTimeout = 30m

pkg/status/config.go
ShettyGaurav: No errors are swallowed — checkDependencies ([status.go:214](vscode-webview://1sfe4u5v568oc29i7jllk0ldqrqs2mi3dc3fcfa4fnlut22488rt/pkg/status/status.go#L214)) has no error return. It handles errors internally (logs to stderr, returns nil slice).

pkg/status/status.go
ShettyGaurav: added enumValue implementing pflag.Value with Set()/Type()/String(). Invalid values are now rejected at parse time, and --help shows --wait-for healthy as the type. Also kept AnnotationValidValues for the API manifest. Removed the manual slices.Contains check from ValidateWait.

pkg/status/status.go
ShettyGaurav: Replaced with healthConfig *clusterhealth.Config

pkg/status/config.go
ShettyGaurav: Removed both wrappers. clusterhealth.Run errors now pass through as-is, which also avoids breaking IsUnrecoverableError's errors.As chain traversal.

macgregor (approved):

dayakar349 (approved):

ShettyGaurav:
@coderabbitai review

PR #114, opened 2026-07-02
feat(watch): add --watch flag with centralized WatchOptions

ShettyGaurav:
### Jira:  
[RHOAIENG-66868](https://redhat.atlassian.net/browse/RHOAIENG-66868)

## Description

Add `--watch` flag to `kubectl odh status` for streaming status changes as NDJSON (`-o json`) or multi-document YAML (`-o yaml`). Only state changes are emitted; duplicates are suppressed.

Watch infrastructure is centralized in `pkg/cmd/WatchOptions` — any command can add `--watch` by embedding it and providing a poller + emitter, same pattern as `WaitOptions`.

```bash
kubectl odh status --watch -o json
kubectl odh status --watch -o json --poll-interval=30s
```

- Mutually exclusive with `--wait-for`; requires `-o json` or `-o yaml`
- Exponential backoff on transient errors; unrecoverable errors stop immediately

## Testing

## Review checklist

## Summary

* **New Features**
  * Added reusable `--watch` streaming for JSON/YAML and integrated watch mode into the `status` command.
  * Watch emits an initial update immediately, then outputs only when the reported state changes.

* **Bug Fixes**
  * Improved CLI validation (structured formats only) with clear mutual exclusivity against wait mode.
  * Enhanced transient watch retries with backoff and warning output, and improved streaming output formatting for JSON (NDJSON) and YAML (multi-document).

* **Tests**
  * Added unit tests covering watch validation, streaming format, duplicate suppression, and retry/state-change behavior.

[RHOAIENG-66868]: https://redhat.atlassian.net/browse/RHOAIENG-66868?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -71,6 +71,7 @@ const (
 // Command contains the status command configuration.
 type Command struct {
 	cmd.WaitOptions
+	cmd.WatchOptions
 
 	IO          iostreams.Interface
 	ConfigFlags *genericclioptions.ConfigFlags
@@ -90,6 +91,10 @@ type Command struct {
 
 	QPS   float32
 	Burst int
+
+	// TimeoutExplicit is true when --timeout was explicitly passed by the user.
+	// When false and --watch is active, timeout defaults to 0 (indefinite).
+	TimeoutExplicit bool
 
 	// Populated during Complete
 	restConfig *rest.Config
@@ -134,6 +139,7 @@ func (c *Command) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&c.IncludeInfra, "include-infra", false, flagDescInfra)
 	fs.BoolVar(&c.IncludeDeps, "include-deps", c.IncludeDeps, flagDescIncludeDeps)
 	c.AddWaitFlags(fs, []string{WaitConditionHealthy})
+	c.AddWatchFlag(fs)
 	fs.Float32Var(&c.QPS, "qps", c.QPS, flagDescQPS)
 	fs.IntVar(&c.Burst, "burst", c.Burst, flagDescBurst)
 }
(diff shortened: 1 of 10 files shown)


pkg/status/status.go
```diff
@@ -91,7 +92,8 @@ type Command struct {
 	QPS   float32
 	Burst int
 
-	// Populated during Complete
+	// Populated during AddFlags/Complete
+	flags      *pflag.FlagSet
```
dayakar349: flags.Changed("timeout") is idiomatic pflag, but tests construct Command without AddFlags, so c.flags is always nil, silently forcing timeout=0. A timeoutExplicit bool field would make the contract explicit and allow future tests to exercise timeout behavior without wiring up a full pflag.FlagSet.

pkg/status/watch_internal_test.go
```diff
@@ -0,0 +1,176 @@
...
+	})
+
+	t.Run("emits on state change", func(t *testing.T) {
+		g := NewWithT(t)
+
+		cfg := unhealthyConfig(scheme)
+		cmd, stdout, _ := newTestWatchCommand(cfg)
+		cmd.Timeout = 0
+
+		dep := testOperatorDeployment()
```
dayakar349, line 95: The "emits on state change" test creates a deployment at ~1.5s and cancels at 4s. If the envtest apiserver is slow to persist, the next poll can miss the change within that window. A channel signaled from the Poller on detection would allow deterministic cancellation instead of relying on wall-clock timing.

pkg/status/status.go
ShettyGaurav: replaced with TimeoutExplicit bool

pkg/status/watch_internal_test.go
ShettyGaurav: test now signals via a channel on each emission

dayakar349 (approved):

asanzgom (approved):

PR #145, opened 2026-08-20
feat(lint): add dashboard upgrade checks for 2.25 → 3.5 path

liday-rh:
## Summary

Add the five dashboard lint checks for the RHOAI 2.25 → 3.5 upgrade path, covering the sidecar-to-standalone migration gaps enumerated in [RHOAIENG-82359](https://issues.redhat.com/browse/RHOAIENG-82359).

| Check | Impact | What it does |
|---|---|---|
| `components.dashboard.resource-capacity` | Blocking | Sums the dashboard pod's container requests and blocks if no single node has enough allocatable CPU/memory to fit it |
| `components.dashboard.rollout-strategy` | Advisory | Warns when the effective `maxUnavailable` resolves to 0, so the rollout can only progress by surging extra pods |
| `components.dashboard.auth-model-migration` | Advisory | Detects dashboard pods still carrying the `oauth-proxy` sidecar |
| `components.dashboard.route-migration` | Advisory | Detects legacy `rhods-dashboard` / `odh-dashboard` routes whose URL will change |
| `components.dashboard.config-compatibility` | Advisory | Verifies `OdhDashboardConfig` exists and reports its version annotation |

All checks are gated by `IsUpgradeFrom2xTo3x` and designed to run **after** the DSC/DSCI CRDs have been updated with the v2 API version (per the Jira constraint).

## Scope

Earlier revisions of this PR also carried three platform CRD checks (webhook availability, `storedVersions` compatibility, DSC/DSCI finalizer health). Those are platform-level rather than dashboard-specific and have been removed, along with `pkg/lint/checks/platform/crd/`.

Rollout strategy was previously a second condition emitted by the resource-capacity check. It is now its own check, matching the Jira breakdown and making each of the five independently reportable.

## Notes on correctness

A few behaviours are worth calling out, since they differ from a naive reading of the pod spec:

- **Init containers.** The pod request the scheduler uses is not the plain container sum. Init containers run to completion first, so per resource the request is the larger of the regular-container sum and the biggest single init container. Native sidecars (init containers with `restartPolicy: Always`) keep running alongside, so they add to the sum instead.
- **Rollout fenceposts.** `maxUnavailable` and `maxSurge` resolve against the real Kubernetes defaults (`25%` each), so omitting the field and setting `25%` explicitly are reported identically. `maxUnavailable` rounds down and `maxSurge` rounds up, matching the scheduler. When both resolve to 0, Kubernetes' own `ResolveFenceposts` clamps `maxUnavailable` to 1, so that combination cannot stall and is not reported. The `Recreate` strategy short-circuits.

## Test plan

- [x] `make build` — compiles without errors
- [x] `make lint` — 0 linter violations
- [x] `go test ./...` — full suite passes, no regressions
- [x] Manual: `./bin/kubectl-odh lint --target-version 3.5.0 -v` against a fresh RHOAI 2.25.10 ROSA cluster — the checks appear and fire correctly:
  - Auth model migration: WARNING (2 pods with oauth-proxy detected)
  - Route migration: WARNING (legacy `rhods-dashboard` route detected)
  - Resource capacity: WARNING (3→9 containers, no autoscaler)
  - Config compatibility: PASS (odh-dashboard-config found, version 2.25.10)

  lint output against test cluster (taken before the CRD checks were removed and rollout strategy was split out):
  image

Jira: [RHOAIENG-82359](https://issues.redhat.com/browse/RHOAIENG-82359)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[RHOAIENG-82359]: https://redhat.atlassian.net/browse/RHOAIENG-82359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHOAIENG-82359]: https://redhat.atlassian.net/browse/RHOAIENG-82359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added upgrade checks for dashboard authentication, configuration compatibility, legacy routes, and rollout strategies during 2.x-to-3.x migrations.
  * Dashboard route diagnostics now identify all legacy routes requiring migration.
  * Added resource definitions for dashboard configuration, cluster autoscaling, and nodes.

* **Bug Fixes**
  * Improved dashboard capacity calculations for initialization containers, native sidecars, and fractional CPU requests.
  * Added clearer insufficient-capacity reporting.

* **Refactor**
  * Removed obsolete platform CRD checks and replaced them with dashboard rollout strategy validation.


liday-rh:
@coderabbit help

liday-rh:
@coderabbit review

pkg/lint/checks/components/dashboard/auth_model_migration.go
```diff
@@ -0,0 +1,135 @@
...
+
+		return nil
+	}
+
+	req.Result.SetCondition(check.NewCondition(
+		conditionTypeAuthModel,
+		metav1.ConditionTrue,
+		check.WithReason(check.ReasonRequirementsMet),
+		check.WithMessage(msgAuthModelMigrated),
+	))
```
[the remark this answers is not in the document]
liday-rh: Fixed in 564d0d7. Changed the message to `"No oauth-proxy sidecar found on dashboard pods"` — it now only states what was actually verified.

pkg/lint/checks/components/dashboard/config_compatibility.go
```diff
@@ -0,0 +1,109 @@
...
+	if config == nil {
+		req.Result.SetCondition(check.NewCondition(
+			conditionTypeConfigCompatibility,
+			metav1.ConditionTrue,
+			check.WithReason(check.ReasonRequirementsMet),
+			check.WithMessage("OdhDashboardConfig access restricted - unable to verify"),
+		))
+
+		return nil
+	}
```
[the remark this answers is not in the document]
liday-rh: Fixed in 564d0d7. The `config == nil` path now reports `ConditionFalse` / `ReasonInsufficientData` / `ImpactAdvisory` instead of claiming `RequirementsMet`.

pkg/lint/checks/components/dashboard/resource_capacity.go
```diff
@@ -0,0 +1,431 @@
...
+		return nil
+	}
+
+	nodes, err := getNodeAllocatable(ctx, req.Client)
+	if err != nil {
+		return err
+	}
+
+	autoscalers, autoErr := req.Client.List(ctx, resources.ClusterAutoscaler)
+	hasAutoscaler := autoErr == nil && len(autoscalers) > 0
```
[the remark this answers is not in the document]
liday-rh: Fixed in 564d0d7. `hasAutoscaler` is now a `*bool` — `nil` when the list call fails, so the advisory is only emitted when we successfully list zero ClusterAutoscaler resources. A list error (missing CRD, RBAC denial) falls through to the pass condition.

pkg/lint/checks/components/dashboard/resource_capacity.go
```diff
@@ -0,0 +1,431 @@
...
+func getDeploymentPodResources(
+	ctx context.Context,
+	cl client.Reader,
+	namespace string,
+) (resourceLookupResult, error) {
+	deploy, err := cl.GetResource(ctx, resources.Deployment, dashboardDeploymentName,
+		client.InNamespace(namespace))
+	if err != nil || deploy == nil {
+		return resourceLookupResult{}, nil //nolint:nilerr // Not-found is expected
+	}
```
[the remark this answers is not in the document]
liday-rh: Fixed in 564d0d7. All three locations addressed:

- **Deployment lookup (L194-198)**: Now returns the error unless `apierrors.IsNotFound(err)` is true; `deploy == nil` handled separately.
- **Rollout deadlock (L334-342)**: No longer fetches the deployment independently — receives the already-fetched deployment from `getDeploymentPodResources`, eliminating both the duplicate API call and the error-swallowing.
- **ClusterAutoscaler (L110-111)**: Covered by the `*bool` change above.

pkg/lint/checks/components/dashboard/resource_capacity.go
```diff
@@ -0,0 +1,431 @@
...
+
+		if req.cpuMillis > maxReq.cpuMillis || req.memoryBytes > maxReq.memoryBytes {
+			maxReq = req
+			maxCount = count
+		}
+	}
+
+	found := maxReq.cpuMillis > 0 || maxReq.memoryBytes > 0
+
+	return resourceLookupResult{resources: maxReq, containerCount: maxCount, found: found}, nil
```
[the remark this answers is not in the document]
liday-rh: Fixed in 564d0d7. CPU and memory are now tracked independently per dimension. Added `TestResourceCapacityCheck_FallbackPodsDimensionMax` which verifies that pod A with 4000m CPU and pod B with 8Gi memory correctly produces a blocking result when no single node can fit both.

pkg/lint/checks/components/dashboard/resource_capacity.go
```diff
@@ -0,0 +1,431 @@
...
+			if q, pErr := resource.ParseQuantity(cpuStr); pErr == nil {
+				total.cpuMillis += q.MilliValue()
+			}
+		}
+
+		if memStr, ok := requests["memory"].(string); ok {
+			if q, pErr := resource.ParseQuantity(memStr); pErr == nil {
+				total.memoryBytes += q.Value()
+			}
+		}
```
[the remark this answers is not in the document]
liday-rh: Fixed in 564d0d7. Added `parseQuantityAny` helper that handles `string`, `int64`, and `float64` types. Used for both container request parsing and node allocatable parsing. Added `TestResourceCapacityCheck_IntegerQuantities` that verifies a deployment with `int64` quantities (`cpu: int64(1)`, `memory: int64(1073741824)`) parses correctly.

pkg/lint/checks/platform/crd/finalizer_orphan.go
```diff
@@ -0,0 +1,156 @@
...
+			check.WithMessage(msgNoFinalizersFound),
+		))
+
+		return dr, nil
+	}
+
+	resourceNames := strings.Join(resourcesWithFinalizers, ", ")
+
+	deployments, err := target.Client.List(ctx, resources.Deployment,
+		client.WithLabelSelector(operatorLabelSelector))
```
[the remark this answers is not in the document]
liday-rh: Fixed in 564d0d7. Added `client.WithNamespace("redhat-ods-operator")` to scope the deployment lookup. The existing test fixture already sets `"namespace": "redhat-ods-operator"` on the operator deployment, so the test continues to pass.

pkg/lint/checks/platform/crd/webhook_availability.go
```diff
@@ -0,0 +1,132 @@
...
+		_, err = target.Client.GetResource(ctx, resources.Service, svcName, client.InNamespace(svcNamespace))
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				issues = append(issues, fmt.Sprintf(msgWebhookServiceMissing, crdName, svcNamespace, svcName))
+
+				continue
+			}
+
+			return nil, fmt.Errorf("checking webhook service %s/%s: %w", svcNamespace, svcName, err)
+		}
```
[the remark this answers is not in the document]
liday-rh: Acknowledged — this is a valid improvement but skipping for now. The `Endpoints` resource type does not exist in the `resources` package yet, so implementing this would require adding the resource definition and plumbing it through the fake client test infrastructure. The current Service existence check catches the most common failure mode (webhook service not deployed). Will track as a follow-up.

liday-rh:
## CodeRabbit review fixes (564d0d7)

All actionable findings addressed:

| Finding | Resolution |
|---|---|
| auth_model: success message claims kube-rbac-proxy | Changed to "No oauth-proxy sidecar found on dashboard pods" |
| config_compatibility: nil config reports RequirementsMet | Now reports `ConditionFalse` / `InsufficientData` / `ImpactAdvisory` |
| resource_capacity: autoscaler list error → "no autoscaler" | `hasAutoscaler` is `*bool`; advisory only on confirmed empty list |
| resource_capacity: read errors collapsed into "absent" | Deployment: propagates non-NotFound errors. Rollout: receives already-fetched deploy (also fixes duplicate API call nitpick). Autoscaler: `*bool` |
| resource_capacity: pod fallback loses per-dimension max | CPU and memory tracked independently + new test |
| resource_capacity: integer quantities silently dropped | Added `parseQuantityAny` for string/int64/float64 + new test |
| finalizer_orphan: deployment lookup not namespace-scoped | Added `client.WithNamespace("redhat-ods-operator")` |

**Skipped:** webhook endpoint readiness check — `Endpoints` resource type doesn't exist in the codebase yet; tracked as follow-up.

**Nitpick — duplicate deployment fetch:** Resolved by passing the deployment from `getDeploymentPodResources` into `checkRolloutDeadlock`.

**Nitpick — test fixture gaps:** Added `TestResourceCapacityCheck_IntegerQuantities` (int64 quantities) and `TestResourceCapacityCheck_FallbackPodsDimensionMax` (cross-dimension max).

liday-rh:
@CodeRabbit review

pkg/lint/checks/components/dashboard/auth_model_migration.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/opendatahub-io/odh-cli/pkg/resources"
+	"github.com/opendatahub-io/odh-cli/pkg/util/client"
+	"github.com/opendatahub-io/odh-cli/pkg/util/jq"
+	"github.com/opendatahub-io/odh-cli/pkg/util/version"
+)
+
+const (
+	containerOAuthProxy    = "oauth-proxy"
+	containerKubeRBACProxy = "kube-rbac-proxy"
+
```
chambridge, line 22: Dead code: `containerKubeRBACProxy` is defined but never referenced anywhere. Remove it, or if there's a plan to use it (e.g. checking that kube-rbac-proxy *is* present after migration), add that logic now.

pkg/lint/checks/platform/crd/webhook_availability.go
```diff
@@ -0,0 +1,132 @@
...
+	"dscinitializations.dscinitialization.opendatahub.io",
+}
+
+const (
+	conditionTypeWebhookHealthy = "WebhookHealthy"
+
+	msgWebhookHealthy        = "All CRD conversion webhooks are healthy"
+	msgWebhookServiceMissing = "CRD %s has conversion webhook strategy but the referenced service %s/%s is missing"
+	msgWebhookNotConfigured  = "No CRDs with conversion webhook strategy found"
+)
```
chambridge: Dead code: `msgWebhookNotConfigured` is never used — `Validate()` either returns issues or falls through to `msgWebhookHealthy`. Remove it.

pkg/lint/checks/platform/crd/finalizer_orphan.go
```diff
@@ -0,0 +1,157 @@
...
+		check.WithReason(check.ReasonRequirementsMet),
+		check.WithMessage(msgFinalizersHealthy),
+	))
+
+	return dr, nil
+}
+
+func isDeploymentHealthy(deploy *unstructured.Unstructured) bool {
+	available, _ := jq.Query[int64](deploy, ".status.availableReplicas")
+
```
chambridge: Type mismatch: jq/gojq always returns `float64` for JSON numbers. `Query[int64]` will fail the direct type assertion and fall through to a JSON marshal/unmarshal roundtrip every time. Works, but wasteful and inconsistent with `resource_capacity.go:357` which correctly uses `Query[float64]` for the same kind of field.

Same issue at lines 153-154.

```go
available, _ := jq.Query[float64](deploy, ".status.availableReplicas")
return available > 0
```

pkg/lint/checks/components/dashboard/resource_capacity.go
```diff
@@ -0,0 +1,459 @@
...
+		metav1.ConditionTrue,
+		check.WithReason(check.ReasonRequirementsMet),
+		check.WithMessage(msgRolloutOK),
+	))
+
+	return nil
+}
+
+func computeEffectiveMaxUnavailable(raw any, replicas int) int {
+	switch v := raw.(type) {
```
chambridge: Nil `raw` (field absent) returns `1` here, but Kubernetes defaults `maxUnavailable` to `25%`. For 2 replicas that's `floor(0.5) = 0`. So explicitly setting `"25%"` triggers a deadlock warning, but omitting the field entirely (same effective value) does not.

Kubernetes prevents actual deadlock via `ResolveFenceposts` clamping, so this isn't a false negative in practice — but worth a comment explaining the deliberate choice, or matching the real K8s default.

pkg/lint/checks/components/dashboard/route_migration.go
```diff
@@ -0,0 +1,99 @@
...
+		client.WithNamespace(req.ApplicationsNamespace))
+	if err != nil {
+		return fmt.Errorf("listing routes: %w", err)
+	}
+
+	for _, route := range routes {
+		name := route.GetName()
+
+		if slices.Contains(legacyDashboardRouteNames, name) {
+			req.Result.SetCondition(check.NewCondition(
```
chambridge: This returns after the first match from `legacyDashboardRouteNames`. If both `rhods-dashboard` and `odh-dashboard` exist, only the first one encountered gets reported. Is that intentional? If so a comment would help; if not, consider collecting all matches.

pkg/lint/checks/components/dashboard/resource_capacity.go
```diff
@@ -0,0 +1,459 @@
...
+			metav1.ConditionTrue,
+			check.WithReason(check.ReasonRequirementsMet),
+			check.WithMessage(msgRolloutOK),
+		))
+
+		return nil
+	}
+
+	maxUnavailRaw, _ := jq.Query[any](deploy, ".spec.strategy.rollingUpdate.maxUnavailable")
+
```
chambridge: Minor: this warns when `maxUnavailable` rounds to 0, but doesn't consider `maxSurge`. If `maxSurge > 0` (K8s default is 25% → 1 for 2 replicas), the rollout makes progress via surge even with `maxUnavailable=0`. The message text ("if new pods cannot schedule alongside old pods") is technically accurate but could confuse operators who know surge is configured. Consider checking both, or noting surge in the message.

pkg/lint/checks/platform/crd/stored_versions.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+	}
+
+	if len(stale) > 0 {
+		return fmt.Sprintf(msgStoredVersionsMismatch,
+			crdName, strings.Join(stale, ", "), strings.Join(servedNames, ", ")), nil
+	}
+
+	return "ok", nil
+}
```
chambridge: Nit: using the magic string `"ok"` as a sentinel is a bit fragile — easy to typo or miss in a future change. A small struct return like `struct{ found bool; issue string }` or a `(found bool, issue string, err error)` tuple would be more self-documenting.

pkg/lint/checks/platform/crd/finalizer_orphan.go
```diff
@@ -0,0 +1,157 @@
...
+func (c *FinalizerOrphanCheck) Validate(
+	ctx context.Context,
+	target check.Target,
+) (*result.DiagnosticResult, error) {
+	dr := c.NewResult()
+
+	if target.TargetVersion != nil {
+		dr.Annotations[check.AnnotationCheckTargetVersion] = target.TargetVersion.String()
+	}
```
chambridge: The manual `dr.Annotations[check.AnnotationCheckTargetVersion]` pattern is copy-pasted across all 3 CRD checks (`finalizer_orphan.go:63`, `stored_versions.go:64-66`, `webhook_availability.go:64-66`). The dashboard checks avoid this via the `validate.Component` builder. A small `validate.Platform()` or `validate.CRD()` builder (or even a shared helper that creates the result and sets the annotation) would remove the duplication.

chambridge (changes requested):
Solid set of checks, well-tested. A few things to clean up — mostly dead code and a type consistency issue.

8 findings: 2 dead code, 4 correctness/consistency, 1 code style, 1 completeness.

pkg/lint/checks/platform/crd/webhook_availability.go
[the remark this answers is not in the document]
liday-rh: Moot now — this PR has been narrowed to the five dashboard checks enumerated in [RHOAIENG-82359](https://issues.redhat.com/browse/RHOAIENG-82359), and the three platform CRD checks were platform-level rather than dashboard-specific. `pkg/lint/checks/platform/crd/` is deleted as of a9aadec, so this file no longer exists. Resolving.

Happy to carry the point forward if these checks land in a follow-up PR.

pkg/lint/checks/platform/crd/webhook_availability.go
liday-rh: Moot now — this PR has been narrowed to the five dashboard checks enumerated in [RHOAIENG-82359](https://issues.redhat.com/browse/RHOAIENG-82359), and the three platform CRD checks were platform-level rather than dashboard-specific. `pkg/lint/checks/platform/crd/` is deleted as of a9aadec, so this file no longer exists. Resolving.

Happy to carry the point forward if these checks land in a follow-up PR.

pkg/lint/checks/platform/crd/finalizer_orphan.go
liday-rh: Moot now — this PR has been narrowed to the five dashboard checks enumerated in [RHOAIENG-82359](https://issues.redhat.com/browse/RHOAIENG-82359), and the three platform CRD checks were platform-level rather than dashboard-specific. `pkg/lint/checks/platform/crd/` is deleted as of a9aadec, so this file no longer exists. Resolving.

Happy to carry the point forward if these checks land in a follow-up PR.

pkg/lint/checks/platform/crd/stored_versions.go
liday-rh: Moot now — this PR has been narrowed to the five dashboard checks enumerated in [RHOAIENG-82359](https://issues.redhat.com/browse/RHOAIENG-82359), and the three platform CRD checks were platform-level rather than dashboard-specific. `pkg/lint/checks/platform/crd/` is deleted as of a9aadec, so this file no longer exists. Resolving.

Happy to carry the point forward if these checks land in a follow-up PR.

pkg/lint/checks/platform/crd/finalizer_orphan.go
liday-rh: Moot now — this PR has been narrowed to the five dashboard checks enumerated in [RHOAIENG-82359](https://issues.redhat.com/browse/RHOAIENG-82359), and the three platform CRD checks were platform-level rather than dashboard-specific. `pkg/lint/checks/platform/crd/` is deleted as of a9aadec, so this file no longer exists. Resolving.

Happy to carry the point forward if these checks land in a follow-up PR.

pkg/lint/checks/components/dashboard/auth_model_migration.go
liday-rh: Removed in a9aadec. There was no plan to use it — the earlier thread on the success message settled on stating only that no `oauth-proxy` sidecar was found (rather than asserting `kube-rbac-proxy` is present), which left the constant orphaned.

pkg/lint/checks/components/dashboard/resource_capacity.go
liday-rh: Good catch, fixed in a9aadec — the inconsistency was real, so I matched the actual Kubernetes defaults rather than papering over it with a comment.

`resolveFencepost` now substitutes `25%` for an absent `maxUnavailable` (and `25%` for an absent `maxSurge`), so omitting the field and setting `25%` explicitly resolve identically and are reported identically. When the value comes from the default, the message says so explicitly: `maxUnavailable=25% (Kubernetes default)`.

Your point about `ResolveFenceposts` clamping is handled too — see the `maxSurge` thread.

pkg/lint/checks/components/dashboard/resource_capacity.go
liday-rh: Fixed in a9aadec. Both fenceposts are now resolved, with the same rounding Kubernetes uses (`maxUnavailable` floors, `maxSurge` ceils), which changed the check's behaviour in two ways:

- When both resolve to 0, Kubernetes' own `ResolveFenceposts` clamps `maxUnavailable` to 1, so no stall is possible and nothing is reported.
- The `Recreate` strategy short-circuits, since neither fencepost applies.

What's left is the case you describe — `maxUnavailable=0` with surge available — which is still worth flagging, because a surge-only rollout progresses only if the cluster can schedule the extra pods, and this check fires on an upgrade where the pod grows from 3 to 9 containers. The message now names the surge count rather than leaving it implicit:

> Deployment "rhods-dashboard" has 2 replica(s) with maxUnavailable=25%, which resolves to 0 — the rolling update can only progress by surging 1 extra pod(s), and will stall if the cluster cannot schedule them alongside the existing pods

This also moved to its own check, `components.dashboard.rollout-strategy`, per the Jira breakdown.

pkg/lint/checks/components/dashboard/route_migration.go
liday-rh: Not intentional — fixed in a9aadec. It now collects every match, sorts them, and reports them together, since a cluster can carry both names and each is a URL the admin has to update:

> Dashboard route(s) "odh-dashboard", "rhods-dashboard" found - URL will change after upgrade to a gateway-based route

Added `TestRouteMigrationCheck_ReportsEveryLegacyRoute` to cover it.

pkg/lint/checks/components/dashboard/resource_capacity.go
```diff
@@ -0,0 +1,459 @@
...
+	if deploy == nil {
+		return resourceLookupResult{}, nil
+	}
+
+	req, count, err := sumContainerRequests(deploy, ".spec.template.spec.containers")
+	if err != nil {
+		return resourceLookupResult{}, fmt.Errorf("reading deployment container requests: %w", err)
+	}
+
+	return resourceLookupResult{resources: req, containerCount: count, found: true, deployment: deploy}, nil
```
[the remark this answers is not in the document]
liday-rh: Addressed in a9aadec. `sumContainerRequests` is replaced by `effectivePodRequests`, which computes the request the scheduler actually uses: per resource, the larger of the regular-container sum and the biggest single init-container request.

One refinement beyond the suggestion — native sidecars (init containers with `restartPolicy: Always`) keep running alongside the regular containers, so they're added to the sum instead of being treated as a floor.

Three regression tests: `InitContainerExceedsNodeCapacity` (blocks), `InitContainerNotSummed` (a plain init container is not added to the sum), and `NativeSidecarIsSummed`.

pkg/lint/checks/components/dashboard/resource_capacity.go
```diff
@@ -0,0 +1,459 @@
...
+
+		return q, err == nil
+	case int64:
+		return *resource.NewQuantity(v, resource.DecimalSI), true
+	case float64:
+		return *resource.NewQuantity(int64(v), resource.DecimalSI), true
+	default:
+		return resource.Quantity{}, false
+	}
+}
```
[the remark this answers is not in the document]
liday-rh: Addressed in a9aadec. One deviation from the suggested diff: I used `'f'` rather than `'g'` in `strconv.FormatFloat`, so the output never carries exponent notation — a memory value like `1073741824` formats as `1073741824` instead of `1.073741824e+09`.

```go
case float64:
	q, err := resource.ParseQuantity(strconv.FormatFloat(v, 'f', -1, 64))

	return q, err == nil
```

Covered by `TestResourceCapacityCheck_FractionalCPUNotTruncated`, which asserts a 1.5 CPU request is reported as `1500m` rather than being truncated to `1`.

liday-rh:
@CodeRabbit review

liday-rh:
@CodeRabbit full review

chambridge (approved):
/lgtm

PR #151, opened 2026-08-31
fix: detect notebooks missing auth migration

wes-spinks:
## Summary

- treat a Notebook without `inject-auth: "true"` as requiring the auth-model patch, even when it has no legacy OAuth artifacts
- prevent `workbenches.patch-auth-model` from reporting such Notebooks as already patched
- add regression coverage for a minimal Notebook without legacy OAuth artifacts or the 3.x auth annotation

## Testing

- `go test -v ./pkg/migrate/actions/workbenches/authmodel -run '^TestNeedsAuthModelPatch_'`
- `make check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated notebook migration checks to identify notebooks missing the required authentication injection annotation.
  * Ensured notebooks without the 3.x authentication configuration are correctly flagged for an authentication-model patch.
  * Continued detecting legacy OAuth artifacts that require migration.


chambridge (approved):
lgtm

